PR #48, opened 2020-02-24
BBBSL-36: Added MAX_MEETING_DURATION env variable that overrites duration

farhatahmad:


farhatahmad:
@kepstin The main issue was users lingering in sessions causing meetings to drag out for a long period of time. It doesn't seem like the Front End offers any way to set this param on create call

PR #50, opened 2020-02-26
BBBSL-31: Import Recordings

kepstin:


Dockerfile
```diff
@@ -31,20 +59,24 @@ RUN bundle config build.nokogiri --user-system-libraries \
     && rm -rf vendor/bundle/ruby/*/cache \
     && find vendor/bundle/ruby/*/gems/ \( -name '*.c' -o -name '*.o' \) -delete
 COPY --chown=scalelite:scalelite . ./
+RUN rm -rf nginx
 
 FROM base AS application
 USER scalelite:scalelite
-ENV RAILS_ENV=production RAILS_LOG_TO_STDOUT=1
+ENV URL_HOST=localhost RAILS_ENV=production RAILS_LOG_TO_STDOUT=1
```
farhatahmad: Is this supposed to be hardcoded to localhost?

kepstin: This is not hardcoded - environment variables set in the Dockerfile are defaults and can be overridden when the container is run.

PR #60, opened 2020-03-05
BBBSL-41: Implemented delete recordings

farhatahmad:


app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -299,6 +299,27 @@ def update_recordings
...
   end
 
+  def delete_recordings
+    raise BBBError.new('missingParamRecordID', 'You must specify a recordID.') if params[:recordID].blank?
+
+    Recording.transaction do
+      query = Recording.where(record_id: params[:recordID].split(','))
+      raise BBBError.new('notFound', 'We could not find recordings') if query.none?
+
+      query.each do |rec|
```
kepstin: Right now the `query.none?` and `query.each` will do two separate database calls. You can fix this by manually indicating when to load the data: Add `.load` to the end of the `query = ` line.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -299,6 +299,27 @@ def update_recordings
...
+        logger.debug("Deleting recording: #{rec.record_id}")
+        # TODO: check the unpublished dir when it is implemented
+        FileUtils.rm_r(Dir.glob(File.join(Rails.configuration.x.recording_publish_dir, '/*/', rec.record_id)))
+        rec.destroy!
+      rescue StandardError => e
+        logger.warn("Error #{e} deleting recording #{rec.record_id}")
+        raise InternalError, 'Unable to delete recording.'
```
kepstin: We'll have actually removed the recording files for some recordings at this point, but raising the error here will abort the transaction, and the deletion of the corresponding recordings in the database will be cancelled.

Maybe we should do a separate transaction per recording.

PR #75, opened 2020-03-17
BBBSL-48: Added custom health checks

farhatahmad:


app/controllers/health_check_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    render(plain: response)
+  end
+
+  private
+
+  def cache_check
+    raise 'Unable to write to cache' unless Rails.cache.write('__health_check_cache_write__', 'true', expires_in: @cache_expire)
+
+    raise 'Unable to read from cache' unless Rails.cache.read('__health_check_cache_write__') == 'true'
```
jfederico: Why this one does not have expiration like the prevous one?

And do we need to compare to String 'true'

app/controllers/health_check_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+class HealthCheckController < ApplicationController
+  def all
+    response = 'success'
+    @cache_expire = 10.seconds
+
+    begin
+      cache_check
+      database_check
+    rescue StandardError => e
+      response = "Health Check Failure: #{e}"
```
jfederico: I would make this the exit point directly, less spagety
and remove the initialization at the top and  just send the constant at the bottom. Plain and simple

jfederico:
It doesn't work. The healt_check should be out of the constrain of URL_HOST

app/controllers/health_check_controller.rb
farhatahmad: The expiry is only used for a cache write, not a cache read. 

The cache read just checks the the value that was stored in the cache from the previous step. It can be set to any string as long as the string is the same in the write and the read

app/controllers/health_check_controller.rb
farhatahmad: Wouldn't that require 2 separate return statements?

app/controllers/health_check_controller.rb
jfederico: True, I misread the intention of the test

Do we need to verify read?

app/controllers/health_check_controller.rb
jfederico: Yes, that is correct. The single point of exit recommendation is long gone in favour of simplicity and readability, that is also how you get rid of nedted if and else statements

app/controllers/health_check_controller.rb
kepstin: Is this supposed to check the connection to redis? Note that the redis connection does not use the Rails cache, and the rails cache is just using the default in-process memory mode, so this doesn't really tell anything.

kepstin: If you want to test the redis connection you should do something like
```ruby
RedisStore.with_connection do |redis|
  # run some redis commands here, the 'redis' var is a connection
end
```

PR #106, opened 2020-03-27
BBBSL-53: NGINX now forces ssl enabled websites to use https

### LINKED ISSUE #86: Thumbnail links use wrong protocol (http) when behind proxy (ELB)
I'm running a scalelite instance behind an ELB which has the SSL cert. The thumbnail URLs are coming through as HTTP rather than HTTPS causing my site to be marked as insecure.

### LINKED ISSUE #87: NGINX_SSL=true doesn't redirect http sessions to https.
I'd expect http sessions redirected to https on an SSL-enabled web server. 

My suggestion is to just change the HTTP server block in `/etc/nginx/conf.d/scalelite-ssl.template` to redirect clients to the SSL server...

If you think the server should also be available on HTTP, please at least make it configurable for those of us that think otherwise.

farhatahmad:
fixes #86 
fixes #87


nginx/conf.d/scalelite-ssl.template
```diff
@@ -4,7 +4,9 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
-	include conf.d/scalelite.common;
+	location / {
+    return  301 https://$host$request_uri;
+  }
```
kepstin: Some indentation issues here, try to be consistent with the rest of the file.

nginx/conf.d/scalelite.template
```diff
@@ -4,5 +4,22 @@ server {
...
+		proxy_pass http://scalelite-api:3000;
+
+		proxy_read_timeout 60s;
+		proxy_redirect off;
+
+		proxy_set_header  Host $http_host;
+
+		proxy_set_header  X-Real-IP $remote_addr;
+		proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
+		proxy_set_header  X-Forwarded-Proto $http_x_forwarded_proto;
```
kepstin: Do not unconditionally add the `proxy_set_header  X-Forwarded-Proto $http_x_forwarded_proto;` since it means the proto can be spoofed if scalelite ports are directly accessible. (Admittedly not a *big* deal here since scalelite doesn't use it for anything, but still.)

We should probably have a separate config for "behind an external https-terminating load balancer" which sets the X-Forwarded-Proto header and uses the [nginx realip module](https://nginx.org/en/docs/http/ngx_http_realip_module.html) to pull the original ip from the forwarded request to log and pass to scalelite.

einhirn:
@farhatahmad, you beat me to it - just saw that you already had a pull request in the works when I was about to send one. My method is a little different - I just
`return 301 https://$URL_HOST$request_uri`
unconditionally and directly from the http server block. This also takes care of Host name variants not in the SSL Certificate...

nginx/conf.d/scalelite-ssl.template
einhirn: The original file uses real tab, not space.

nginx/conf.d/scalelite.template
farhatahmad: Is it not be safe to assume that anyone not using SSL would be placing Scalelite behind a network load balancer?

nginx/conf.d/scalelite-ssl.template
jfederico: Well, we should replace tabs by spaces. Let the editors do their job :)

farhatahmad:
Replaced by #139

einhirn:
Hmm... but #139 doesn't change anything in the non-ssl or ssl-template, it just adds a new one...

farhatahmad:
@einhirn Yes you're correct. I'll see if the author of that PR can add the redirect as well

Issue #99, opened 2020-03-25
Bring number of users into the server-load


paul1278:
For example, if you have two servers and there are three meetings running:
```
server1:    1 Meeting, 50 active users
server2:    2 Meetings, 5 active users
```
When now a fourth meeting wants to start, it will be on server1, but it would be better on server2, or did I misunderstand something?
Because the server-load is the number of meetings on a server, as seen here: https://github.com/blindsidenetworks/scalelite/blob/master/lib/tasks/poll.rake#L36 

and here (only incremented on meeting create): 
https://github.com/blindsidenetworks/scalelite/blob/89a658adcdf6729f2a238c8d923e3607a00a18e0/app/controllers/bigbluebutton_api_controller.rb#L149

farhatahmad:
Hi @paul1278,

No you're correct. Scalelite was designed to equal balance meetings across servers based only on the number of meetings. Ideally, there are multiple factors that tie into correctly load balancing across a server. The ones that come to mind are:
- Server weight
- Number of meetings
- Number of users
- Number of streams

Not quite sure if/when this will be changed as out current focus now is mostly on maintenance of our infrastructure and community support

paul1278:
Hi @farhatahmad,

oh ok, thank you for that information! Nevertheless good work!

einhirn:
I'll look into providing a PR with a slight improvement over the current situation. I think it should be doable, since the poller requests "getMeetings" anyway, so I'll only have to dig a little bit deeper into the XML, there's a count of Videos and Audios or something in there...

paul1278:
That would be veeeery nice thank you so much!

paul1278:
Oh @einhirn, what I did on another script (has nothing to do with scalelite):
per Meeting, there is a tag
* `participantCount`, holding the total participants in the meeting
* `listenerCount`, holding the amount of people just listening
* `voiceParticipantCount`, holding the amount of people listening & with microphone turned on
* `videoCount`, holding the amount of people with turned on video

My ruby is not that good either, but if you have any problems, I am happy to help.

einhirn:
Thanks @paul1278, I found out that the "status" task already queries all the fields I want to use anyway, so I just copied that code with minor adjustments 😁

paul1278:
Oh very nice forgot about that! Really appreciate your work! :D

[...]

PR #108, opened 2020-03-29
use video_streams and participants in computing server load

einhirn:
As announced in #99 I had a go at changing the load computation. I borrowed code from the 'status' task and just multiply video_streams by 100, Users by 10 and add everything together with the meeting count. I don't know if there's a better way to compute the load or whether this might lead to an overflow of the variable in some configuration...

Adding a meeting still increases the load by 1 - I don't know whether that's still enough or whether it's needed at all, but it will be overwritten after the next polling cycle anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,9 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `RECORDING_WORK_DIR`: Directory where temporary files from recording transfer/import are extracted. Defaults to `/var/bigbluebutton/recording/scalelite`
 * `RECORDING_PUBLISH_DIR`: Directory where published recording files are placed to make them available to the web server. Defaults to `/var/bigbluebutton/published`
 * `RECORDING_UNPUBLISH_DIR`: Directory where unpublished recording files are placed to make them unavailable to the web server. Defaults to `/var/bigbluebutton/unpublished`
+* `LOAD_WEIGHT_VIDEOS`: Integer factor the number of video streams on a server are multiplied with in load calculation. Defaults to 100.
+* `LOAD_WEIGHT_USERS`: Integer factor the number of video streams on a server are multiplied with in load calculation. Defaults to 10.
+* `LOAD_WEIGHT_MEETINGS`: Integer factor the number of video streams on a server are multiplied with in load calculation. Defaults to 1.
 
 ### Redis Connection (`config/redis_store.yml`)
 
--- a/lib/tasks/poll.rake
+++ b/lib/tasks/poll.rake
@@ -33,7 +33,20 @@ namespace :poll do
       Rails.logger.debug("Polling Server id=#{server.id}")
       resp = get_post_req(encode_bbb_uri('getMeetings', server.url, server.secret))
       meetings = resp.xpath('/response/meetings/meeting')
-      server.load = meetings.length
+
+      server_users = 0
+      video_streams = 0
+
+      meetings.each do |meeting|
+        count = meeting.at_xpath('participantCount')
+        users = count.present? ? count.text.to_i : 0
+        server_users += users
+
+        streams = meeting.at_xpath('videoCount')
+        video_streams += streams.present? ? streams.text.to_i : 0
+      end
+
+      server.load = video_streams * 100 + server_users * 10 + meetings.length
       server.online = true
     rescue StandardError => e
       Rails.logger.warn("Failed to get server id=#{server.id} status: #{e}")
--- a/lib/tasks/poll.rake
+++ b/lib/tasks/poll.rake
@@ -28,6 +28,10 @@ namespace :poll do
   task servers: :environment do
     include ApiHelper
 
+    weight_videos = ENV.has_key?('LOAD_WEIGHT_VIDEOS') ? ENV['LOAD_WEIGHT_VIDEOS'].to_i : 100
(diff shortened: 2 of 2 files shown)


farhatahmad:
Hi @einhirn 

We appreciate you putting the work in to get this implemented. We actually have a plan to implement this is the next major release for Scalelite. We still need to do more research and testing to see what the best way to balance the load is, so once we've got that, we'll put together a PR making the changes.

Thanks again

defnull:
We improved upon this idea and added separate load factors for audio/video downstreams (in addition to upstreams). Downstream counts depend on the individual meeting size and must be calculated per meeting, then summed up. But since we are iterating over all meetings anyway, that was easy to add.
Note that audio and video downstreams must be calculated differently, because BBB mixes audio into a single channel, but does not do that for video. A meeting with 10 participants, each transmitting video, has roughly the same video downstream load as a lecture with a single presenter and 100 viewers.

Since this pull request was rejected, I assume that a new one will also be rejected. Source is available on request as per AGPL, if anyone is interested.

einhirn:
@defnull please mention this also in #99, @farhatahmad revisited it about 5 days ago, maybe your improvement will help...

rabser:
I'm interested to test your changes.

defnull:
It boils down to an updated `servers` task in `poll.rake`:

```
  task servers: :environment do
    include ApiHelper

    weight_meetings = ENV['LOAD_WEIGHT_MEETINGS'].to_d || 1.0
    weight_users = ENV['LOAD_WEIGHT_USERS'].to_d || 0
    weight_audio_rx = ENV['LOAD_WEIGHT_AUDIO_RX'].to_d || ENV['LOAD_WEIGHT_AUDIO'].to_d || 0
    weight_audio_tx = ENV['LOAD_WEIGHT_AUDIO_TX'].to_d || ENV['LOAD_WEIGHT_AUDIO'].to_d || 0
    weight_video_rx = ENV['LOAD_WEIGHT_VIDEO_RX'].to_d || ENV['LOAD_WEIGHT_VIDEO'].to_d || 0
    weight_video_tx = ENV['LOAD_WEIGHT_VIDEO_TX'].to_d || ENV['LOAD_WEIGHT_VIDEO'].to_d || 0

    Rails.logger.debug('Polling servers')
    Server.all.each do |server|
      Rails.logger.debug("Polling Server id=#{server.id}")
      resp = get_post_req(encode_bbb_uri('getMeetings', server.url, server.secret))

      load = 0.0
      resp.xpath('/response/meetings/meeting').each do |meeting|
        users = meeting.at('participantCount')&.content.to_i
        audio = meeting.at('voiceCount')&.content.to_i
        video = meeting.at('videoCount')&.content.to_i
        load += weight_meetings * 1
        load += weight_users * users
        # Audio is mixed server-side -> Only one downstream per user
        load += weight_audio_rx * audio
        load += weight_audio_tx * users if audio
        # Video is NOT mixed server-side -> One downstream per user per video
        load += weight_video_rx * video
        load += weight_video_tx * users * video
      end
      load *= server.load_multiplier.to_d if server.load_multiplier
      server.load = load
      server.online = true
    rescue StandardError => e
      Rails.logger.warn("Failed to get server id=#{server.id} status: #{e}")
      server.load = nil
      server.online = false
    ensure
      begin
        server.save!
        Rails.logger.info(
          "Server id=#{server.id} #{server.online ? 'online' : 'offline'} " \
          "load: #{server.load.nil? ? 'unavailable' : server.load}"
        )
      rescue ApplicationRedisRecord::RecordNotSaved => e
        Rails.logger.warn("Unable to update Server id=#{server.id}: #{e}")
      end
    end
  end
```

The `server.load_multiplier` setting comes from #113.

jodoma:
I like that solution and will give it a try in production. What remains a problem is that multiple LB requests that arrive within a poll interval will all be scheduled to the same server, won't they?

defnull:
Yes, because Scalelite does not know about the additional load factors until it polls all the meetings again. 60 seconds is a short window, thought. I do not think that this is a real issue, compared to another aspect: Lecture-Rooms are usually opened ahead of time. It is entirely possible that a dozen lecture rooms are opened on the same server, now that meeting count is more or less irrelevant. Once people start to join, everything goes up in flames. Hmm perhaps this was not a good idea after all.. :/

einhirn:
Yes - there's a "load + 1" in the handling of a create call. One could change that to something else. Look at https://github.com/blindsidenetworks/scalelite/blob/76105c117c051334277578799c098dfe1313cac4/app/controllers/bigbluebutton_api_controller.rb#L151
I thought about using e.g. a load increment based on an average of ressources currently in use - or you could try to train a neural network to predict when Lecture rooms usually are starting. Ok, just kidding.

einhirn:
> It is entirely possible that a dozen lecture rooms are opened on the same server, now that meeting count is more or less irrelevant.

But the "increase by 1" on create should distribute the new rooms across the servers when they are evenly loaded to begin with.
If there's one nodes that already has a video running (greatly higher load value) it won't be considered. But for all the other nodes this round-robin component should do the trick already...

defnull:
The problem is that "increase by 1" may not actually do anything, depending on the weight factors. It is very unlikely that two servers are only a single load-point apart, if all the other factors are also considered.

defnull:
One idea would be to do a weighted random select, instead of lowest load.

PR #113, opened 2020-03-30
Load-multiplier to weight servers

paul1278:
With this changes a load-multiplier can be added to the servers. It will be multiplicated with the server-load.

Also added a small check for the parameters when adding servers

### DIFF
--- a/lib/tasks/servers.rake
+++ b/lib/tasks/servers.rake
@@ -23,14 +23,12 @@ namespace :servers do
       exit(1)
     end
     tmp_load_multiplier = 1.0
-    if args.load_multiplier.nil?
+    if !args.load_multiplier.nil?
       tmp_load_multiplier = args.load_multiplier.to_d
       if tmp_load_multiplier.zero?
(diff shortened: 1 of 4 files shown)


defnull:
We tested this successfully and are running this patch in production.

paul1278:
@defnull Glad to hear that!

lib/tasks/servers.rake
```diff
@@ -10,16 +10,29 @@ task servers: :environment do
     puts("\tsecret: #{server.secret}")
     puts("\t#{server.enabled ? 'enabled' : 'disabled'}")
     puts("\tload: #{server.load.presence || 'unavailable'}")
+    puts("\tload multiplier: #{server.load_multiplier}")
```
farhatahmad, line 13: Can this be changed to `server.load_multiplier.nil? ? 1.0 : server.load_multiplier.to_d`

paul1278: Done on fd9dcf7

PR #115, opened 2020-04-02
Added support for multiple shared secrets.

defnull:
In environments where multiple applications are integrated with a single
scalelite server, it might be sensible to give each application its own
secret. If an application secret is compromized, it can be revoked
without disturbing any of the other applications.

This patch adds an additional `LOADBALANCER_SECRETS` config knob that may
contain additional secrets, separated by `:`. Any of the configured secrets
will work. The change is fully backwards compatible.


koutzasnew:
In case multiple applications are integrated with a single scalelite server, the URL_HOST in /etc/default/scalelite should be left blank? Tnx

defnull:
Huh? Why? The applications still all access scalelite through the same API URL, they just use individual secrets. The secret is not part of `URL_HOST`.

maths22:
This would also be useful for almost-zero-downtime secret rotation, since a new secret could be added to scalelite and then the api consumers could be updated to use the new secret over time before removing the old secret.

defnull:
We tested this successfully and are running this patch in production.

davidpesce:
@farhatahmad - any chance we can include this in v1.0.2? This is a great security enhancement.

farhatahmad:
@davidpesce Probably not in 1.0.2. Still need to review the code and discuss internally. Assuming everything works out, we'll get it in 1.0.3

README.md
```diff
@@ -110,6 +110,7 @@ To switch your Front-End application to use Scalelite instead of a single BigBlu
 * `URL_HOST`: The hostname that the application API endpoint is accessible from. Used to protect against DNS rebinding attacks. Should be left blank if deploying Scalelite behind a Network Loadbalancer.
 * `SECRET_KEY_BASE`: A secret used internally by Rails. Should be unique per deployment. Generate with `bundle exec rake secret` or `openssl rand -hex 64`.
 * `LOADBALANCER_SECRET`: The shared secret that applications will use when calling BigBlueButton APIs on the load balancer. Generate with `openssl rand -hex 32`
+* `LOADBALANCER_SECRETS`: Additional shared secrets, separated by `:`. Any of these secrets will work. In an environment where multiple applications need to integrate with a single scalelite server, it may be sensible to give each application its own secret. This way, revoking individual secrets later will not disturb other applications.
```
farhatahmad, line 113: Probably should separate them with a `,` and not `:`

defnull:
Is the `:` vs `,` the only reason why this was not included in `v1.0.3`?

README.md
defnull: I opted for `:` because it is the default separator for other important environment variables (e.g. `PATH`) and `,` may imply that more sophisticated csv syntax is allowed (with whitespace stripping, quotes) which is not the case.

farhatahmad:
Nope. I normally only have about ~5 hours a week working on Scalelite which is just enough time to either fix some issues or test + merge a new feature but not both. The past couple of releases were just fixing bugs to get Scalelite as stable as possible. 

Once I can confirm it's a bit more stable, I'll start reviewing PRs, testing and merging as needed

basisbit (approved):
looks good to me.

PR #139, opened 2020-04-09
BBBSL-53: NGINX now forces ssl enabled websites to use https

davidpesce:
Creates a new nginx template and ENV variable for scalelite behind a proxy or load balancer.

### DIFF
--- a/nginx/conf.d/scalelite-ssl.template
+++ b/nginx/conf.d/scalelite-ssl.template
@@ -4,6 +4,10 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
+    location / {
+        return  301 https://$host$request_uri;
+      }
+      
 	include conf.d/scalelite.common;
 }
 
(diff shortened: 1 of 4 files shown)


farhatahmad:
Hi @davidpesce 

Thanks for the PR. There's one part missing, which would be to force an https redirect if ssl is enabled. Can you include it as part of this PR?

This code should be added here: https://github.com/blindsidenetworks/scalelite/blob/master/nginx/conf.d/scalelite-ssl.template#L6
```
location / {
    return  301 https://$host$request_uri;
  }
```

davidpesce:
@farhatahmad - I think this setting would relate to if you had configured scalelite with an actual SSL cert vs being behind a proxy. Right? The new template is scalelite-proxy.template

I made the change anyway because it is necessary with the SSL cert.

davidpesce:
@jfederico - should I resolve the conflict?

farhatahmad:
Hi @davidpesce,

Yes, please resolve the conflict

davidpesce:
@farhatahmad is there a timeline for when v1.0.2 will be released? If today, I'll switch gears and resolve this.

farhatahmad:
@davidpesce Yeah the plan is to release today

davidpesce:
@farhatahmad - all set!

nginx/conf.d/scalelite-ssl.template
```diff
@@ -4,8 +4,11 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
+    location / {
+        return  301 https://$host$request_uri;
+      }
+
 	include conf.d/scalelite/*;
-	include conf.d/scalelite.common;
```
farhatahmad, line 8: Did you mean to remove this?

davidpesce: Yeah, I was following @jfederico 's lead on it. The individual file reference was removed in his latest commit.

farhatahmad: Ah my mistake. I got confused with something else

PR #201, opened 2020-05-09
Run task status in Parallel to connect to servers

rriscado:
If we have a server offline, the task status currently waits for TIMEOUT (10 seconds) to connect in the next server.
If there are a large number of servers offline, it's can take a long time!

These changes I made are to parallelize the execution using threads.


jfederico:
@rriscado thanks for the PR. Although the review may be coming a little overdue, your PR capture my attention because we are looking into this particular issue right now.

I am taking your code and bringing it a different fork because it has some issues that made rubocop mad and they would have to be addressed first.

jfederico:
Discarded in favour of #428

Issue #147, opened 2020-04-11
Configuring the BigBlueButton recording transfer


mohamedelashhab:
I installed scalelite and i have 4 polls , every thing work fine and all polls online and connected but the published records not tranfered to main server scalelite i have a trouble with transfer configuration please help

agustinmono:
FUCKIGN SOLVED THE PROBLEM! AFTER 4 DAYS IT WAS JUST A DUMB CONFIG


This is what I done, no luck making it work, but maybe you can try 
https://vitux.com/install-nfs-server-and-client-on-ubuntu/

Server NFS Side - Scalelite:

 `sudo apt install nfs-kernel-server`
`sudo mkdir -p /mnt/scalelite-recordings`
`sudo chown nobody:nogroup /mnt/scalelite-recordings`
`sudo chmod 777 /mnt/scalelite-recordings`
`sudo nano /etc/exports
`
Inside exports, one line by BBB IP, this is the most important part, in the documentation says ,
(rw,sync,no_subtree_check), insted you need to: 

/mnt/scalelite-recordings Client_IP_BBB(rw,sync,subtree_check)

`sudo exportfs -a`

`sudo systemctl restart nfs-kernel-server`

`sudo ufw allow from [clientIP or clientSubnetIP] to any port nfs`


Client NFS Silde - BigBlueButton

`sudo apt-get update`
`sudo apt-get install nfs-common`
`sudo mkdir -p /mnt/scalelite-recordings`
`sudo mount serverIP:/mnt/scalelite-recordings /mnt/scalelite-recordings`

Jumped to : https://github.com/blindsidenetworks/scalelite/blob/master/sharedvolume-README.md


Now that its working on one server ill try to replicate to my 8 bbb servers, I need to sleep first! 

Good luck

agustinmono:
This erros show upo when i run the script  

`sudo sh scalelite_batch_import.sh`

root@bbb-01:/# sudo sh scalelite_batch_import.sh
scalelite_batch_import.sh: 15: scalelite_batch_import.sh: [[: not found
/usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb:10: warning: constant Gem::ConfigMap is deprecated
Transferring recording for 05a733aeda9e727f47f0d7103486ad8151e332d2-1586463351690 to Scalelite
Found recording format: presentation/05a733aeda9e727f47f0d7103486ad8151e332d2-1586463351690
Creating recording archive
[...]

Issue #129, opened 2020-04-07
Cant transfer old records


agustinmono:
When I run the provided script 

https://github.com/blindsidenetworks/scalelite/blob/master/bigbluebutton/scalelite_batch_import.sh

i have this error
http://prntscr.com/rurrkp

`scalelite_batch_import.sh: 15: scalelite_batch_import.sh: [[: not found
sudo: ./post_publish/scalelite_post_publish.rb: command not found`

IamLuksha:
Change 

`( cd "$scripts_dir" && sudo -n -u bigbluebutton ./post_publish/scalelite_post_publish.rb -m "$record_id" )`

to

`( cd "$scripts_dir" && sudo -n -u bigbluebutton ruby ./post_publish/scalelite_post_publish.rb -m "$record_id" )`

agustinmono:
Done!
 now i get
Permission denied, how can I change that? thanks 
http://prntscr.com/rwvrbe

agustinmono:
> Change
> 
> `( cd "$scripts_dir" && sudo -n -u bigbluebutton ./post_publish/scalelite_post_publish.rb -m "$record_id" )`
> 
> to
> 
> `( cd "$scripts_dir" && sudo -n -u bigbluebutton ruby ./post_publish/scalelite_post_publish.rb -m "$record_id" )`

@kepstin  Can you take  a look?  the main script here has that mistake. Thanks
https://github.com/blindsidenetworks/scalelite/blob/master/bigbluebutton/scalelite_batch_import.sh

IamLuksha:
@agustinmono Have you done everything here?
https://github.com/blindsidenetworks/scalelite/blob/master/sharedvolume-README.md

agustinmono:
Yes @IamLuksha  , I have 2 concerns .

1- I Followed[ this instuctive](https://vitux.com/install-nfs-server-and-client-on-ubuntu/) to create the NFS, configuring the Scalelite Server as NFS Server, and each BBB Server as NFS Clients, its that ok? 


2- here https://github.com/blindsidenetworks/scalelite/blob/master/bigbluebutton/README.md#shared-filesystem-eg-nfs-mount says 

" Ensure that the bigbluebutton user has permission to enter and write to the spool directory."

How can I test that? in case that bigbluebutton doesnt has permission, how can I give permissions to him.

Thanks a lot

IamLuksha:
@agustinmono 
There is an example setup via ssh.
[...]

Issue #83, opened 2020-03-20
Docker-Compose?


einhirn:
Hi, when starting to setup scalelite, I thought that it would have been nice to have a tarball of the systemd service files so I wouldn't need to copy&paste them to my system...

I finally went with docker-compose, so I didn't need the files you provided, but maybe someone else would like that, too.

aakatev:
Off topic, but do you mind sharing docker-compose file?

einhirn:
Not at all... This combines everything needed - the scalelite containers, redis, postgres, greenlight (using 'customize' instructions). I needed to modify the 'greenlight.nginx' file a little (nginx does a longest match for locations, so greenlight's '/b' erroneously matched '/bigbluebutton', too. Adding a trailing slash to the location helped, it works fine now. Oh, and I linked scalelites Environment file to `.env` because of the variables used in Volumes etc.

Using "restart: unless-stopped" makes the containers reappear after a reboot, so I actually don't need any service files. But I lose a bit flexibility to restart e.g. greenlight after a config change without restarting scalelite, too. Reloading the Env file usually takes `docker-compose down` followed by a `... up (-d) ` afterwards - except when you try to `source` the env-file before reloading a container...

```yaml
version: '3'

services:
  greenlight:
    entrypoint: [bin/start]
    image: greenlight-local:release-v2
    container_name: greenlight-v2
    env_file: /etc/default/greenlight
    restart: unless-stopped
    volumes:
      - /var/log/greenlight:/usr/src/app/log
    depends_on:
      - db
  db:
    image: postgres:9.5
    restart: unless-stopped
    volumes:
      - /opt/postgres:/var/lib/postgresql/data
    environment:
      - POSTGRES_DB=postgres
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=very_secret
  redis:
    image: redis:latest
    command: redis-server --appendonly yes
    restart: unless-stopped
    volumes:
      - /opt/redis:/data
  scalelite-api:
    image: blindsidenetwks/scalelite:${SCALELITE_TAG}-api
    container_name: scalelite-api
    restart: unless-stopped
    env_file: /etc/default/scalelite
    volumes:
      - ${SCALELITE_RECORDING_DIR}:/var/bigbluebutton
    depends_on:
      - db
      - redis
  scalelite-nginx:
    image: blindsidenetwks/scalelite:${SCALELITE_TAG}-nginx
    restart: unless-stopped
[...]

PR #203, opened 2020-05-11
Rationalising the UID and GID of bigbluebutton user in Scalelite containers

lightweight:
This pull request covers my tweaks to the Scalelite Dockerfile to set the UID and GID both to 997 as seems to be the default set via the bbb-install script for BBB instances. I have also included a quick script for building an deploying our containers that someone else can easily repurpose. 

** Reasoning **

I have found it difficult (and likely dangerous) to manage recordings on instances of BBB, created using the bbb-install script on Ubuntu 16.04 hosts, sitting behind a combined Scalelite and Greenlight load balancer (running in Docker via Docker-Compose). 

I have succeeded in setting up NFS and having Scalelite see the same recordings file system as my two BBB "workers". But I haven't got the processing of recordings working properly using the existing Scalelite documentation plus insights from #83, #129, #147, and other issues. I noticed that some people had got the recordings working properly.  I've managed to create a *few* published recordings by manually pushing raw recordings through the process via bbb-record and changes of ownership for the files and directories involved, but I have not yet had any recordings "published" without resorting to `sudo -n -u bigbluebutton bash /var/bigbluebutton/scalelite_batch_import.sh`. 

It appears some folks have managed to get it working (I'm not sure how, to be honest - I suspect some undocumented umask fiddling or xattr or similar)... by making the whole /var/bigbluebutton/* world read & writeable... but even that is, I suspect, temporary. 

The main source of problems seems to be the difference in UID, GID of bigbluebutton in the Docker containers (1000, 1000) and on the BBB workers (997, 997). The recommendation in #147 involves using a separate group (this is where, I think, some tricky user mapping or umasks would be necessary) but that seems unnecessarily complicated.

My approach is to set the UID and GIDs for the bigbluebutton user in the Scalelite Docker containers.

** Caveats **

This implementation might be slightly better if it used ENV variables or some other mechanism to abstract the actual UID and GIDs in the event that people are using some other system for implementing the BBB works that doesn't result in the specific UID and GID I've used. But in that case, they'll be no worse off than they are with the default Scalelite Docker containers' choice of 1000, 1000.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,8 @@ COPY nginx /etc/nginx/
 EXPOSE 80
 EXPOSE 443
 ENV NGINX_HOSTNAME=localhost
+ENV GID=997
+ENV UID=997
 CMD [ "/etc/nginx/start", "-g", "daemon off;" ]
 
 FROM alpine AS base
@@ -39,9 +41,9 @@ RUN apk add --no-cache \
     ruby-bundler \
     ruby-json \
     tini \
-    tzdata \
-    && addgroup scalelite \
-    && adduser -h /srv/scalelite -G scalelite -D scalelite
+    tzdata 
+RUN addgroup -S -g 997 scalelite \
+    && adduser -S -G scalelite -u 997 -s /bin/false -h /srv/scalelite scalelite
 WORKDIR /srv/scalelite
 
 FROM base as builder
--- a/README.oeru
+++ b/README.oeru
@@ -1,0 +1,1 @@
+Insight on Blindside Networks' docker process: https://github.com/blindsidenetworks/scalelite/blob/master/.gitlab-ci.yml#L59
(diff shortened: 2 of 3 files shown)


davidpesce:
Hi @lightweight - Take a look at this section of the scalelite documentation - https://github.com/blindsidenetworks/scalelite/blob/master/sharedvolume-README.md#setting-up-directory-structure-and-permissions

When doing these commands be sure to note whether they are performed on each of the bbb servers or on scalelite. Hopefully that helps get you to where you need to be!

lightweight:
Thanks @davidpesce - yes, I did do those, but it still creates an awkward situation, where the files written by, say, a bbb server, is written by the bigbluebutton (UID=997) user, and group (GID=997) rather than the recommended "scalelite-spool" group... in which case things fall apart unless the user/group are changed. People have simply chmodded the dirs to 777 which, of course, isn't particularly desirable. I'm just thinking that all of these complexities can be removed if the UID and GID of the bigbluebutton users in the Scalelite containers is also 997:997 rather than 1000:1000... (that still leaves the Red5 written stuff, but that seems less onerous.

PR #205, opened 2020-05-13
Use Hostname As ID

lkiesow:
When managing a larger cluster and wanting to take down specific hosts
or chunks, is it quite inconvenient to find out the specific IDs handed
out to servers in scalelite. Being able to use the domain names instead
would be much nicer.

While it is conceivable that multiple servers share the same domain
name, this is actually pretty unlikely. Usually, the domain name should
be unique and work just fine as identifier.

This patch switches to using the domain name as an identifier by default.
If required for any reason, people can still switch to using UUIDs
though. Old IDs will, of course, stay valid.

---

Example:

```
% ./bin/rake 'servers' | grep -A1 id:

id: bbb03.lkiesow.io
        url: https://bbb03.lkiesow.io/
--
id: bbb02.lkiesow.io
        url: https://bbb02.lkiesow.io/
--
id: bbb01.lkiesow.io
        url: https://bbb01.lkiesow.io/
```


kepstin:
Closing this in favour of #470

PR #206, opened 2020-05-13
Allow Ruby 2.7

lkiesow:
This patch allows Ruby 2.7 which works just fine although there are a
few deprecation warnings.


Ahmet-Kaplan:
You should change [Travis setting](https://github.com/blindsidenetworks/scalelite/blob/5c60e14d8c3addedf959faf9dc0562ad5fe7a4e1/.travis.yml#L5) with 2.7

lkiesow:
The greater was actually a mistake since I didn't mean to break 2.6 compatibility. That's now fixed. Also, unfortunately, Travis CI does [not yet seem to support 2.7](https://docs.travis-ci.com/user/reference/bionic/#ruby-support)

aguerson:
Hi @lkiesow,

"Also, unfortunately, Travis CI does not yet seem to support 2.7"

So we can migrate scalelite to ruby 2.7 or we can't due to incomptatiblity with the supported version of  Travis CI ?
Regards,

lkiesow:
This patch just allows using 2.7. It doesn't change how/if Scalelite + Ruby 2.6 works.

Travis CI will still test using Ruby 2.6 since 2.7 is not yet available. But the tests will still be run and work in the same way they worked before..

aguerson:
@lkiesow 
I'm not speaking about the tests.
I want to run scaletite with ruby 2.7.1 : Is it possible or not with this patch ?
Regards,

lkiesow:
Yes, it works with 2.6.x as well as with 2.7.x:

> This patch allows Ruby 2.7 which works just fine although there are a
> few deprecation warnings.

```sh
$ bundle exec puma -C config/puma.rb
Ignoring commonmarker-0.17.13 because its extensions are not built. Try: gem pristine commonmarker --version 0.17.13
Ignoring nokogiri-1.10.5 because its extensions are not built. Try: gem pristine nokogiri --version 1.10.5
Puma starting in single mode...
* Version 4.3.3 (ruby 2.7.1-p83), codename: Mysterious Traveller
* Min threads: 5, max threads: 5
* Environment: development
/home/lars/dev/bigbluebutton/scalelite/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.2.1/lib/action_dispatch/middleware/stack.rb:37: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/lars/dev/bigbluebutton/scalelite/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.2.1/lib/action_dispatch/middleware/static.rb:110: warning: The called method `initialize' is defined here
* Listening on tcp://0.0.0.0:3000
Use Ctrl-C to stop
```

```
% curl localhost:3000/bigbluebutton/api
<?xml version="1.0"?>
<response>
  <returncode>SUCCESS</returncode>
  <version>2.0</version>
</response>
```

PR #209, opened 2020-05-15
Increased default hash_bucket_size to 64

### LINKED ISSUE #114: scalelite-nginx service: server_names_hash_bucket_size
I get the error below for the scalelite-nginx.service:

```
Apr 01 20:27:39 sc.aabbccddee.ke docker[7820]: 2020/04/01 17:27:39 [emerg] 7#7: could not build server_names_hash, you should increase server_names_hash_bucket_size: 32
Apr 01 20:27:39 sc.aabbccddee.ke docker[7820]: nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 32
```
The actual length of my domain name is sc.aabbccddee.ke (16char). I know I can edit the image (do you recommend?) and fix this in nginx.conf by adjusting the bucket size upwards.

Can you consider raising this value in the container so that it works for slightly longer domain names without having to edit the image manually? For now I have it working using a short domain name but I would like it work with the longer domain name since it is the one reserved for the project.

farhatahmad:
First 2 files are just formatting fixes

fixes #114

### DIFF
--- a/nginx/conf.d/scalelite-ssl.template
+++ b/nginx/conf.d/scalelite-ssl.template
@@ -4,9 +4,9 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
-    location / {
-        return  301 https://$host$request_uri;
-      }
+	location / {
+		return  301 https://$host$request_uri;
+	}
 
 	include conf.d/scalelite/*;
 }
--- a/nginx/conf.d/scalelite-ssl.template
+++ b/nginx/conf.d/scalelite-ssl.template
@@ -1,35 +1,35 @@
 access_log stderr combined;
 
 server {
-	listen 80 default_server;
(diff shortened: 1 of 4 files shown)


nginx/conf.d/scalelite-ssl.template
```diff
@@ -4,9 +4,9 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
-    location / {
-        return  301 https://$host$request_uri;
-      }
+	location / {
+		return  301 https://$host$request_uri;
```
lkiesow, line 8: This actually changes the formatting in a different manner than before where you used two spaces.

farhatahmad: Should be fixed now

PR #254, opened 2020-06-23
Added a JSON optional format to Server list

rabser:
I needed to parse the servers list, so i added an optional parameter to servers list rake task. If the parameter is actualized with the string "json", you'll get a json formatted list of the servers.
If the parameter is not defined, the servers list is printed out as before.

I'm using it to search a server ID from the hostname:

```
docker exec -it scalelite-api ./bin/rake servers[json] | | jq .servers | jq -c ".[] | select( .url | contains(\"bbbhost1.mydomain.com\"))" | jq -r .id`


defnull:
This is now (more or less) solved with `servers:yaml`. A `servers:json` task could be implemented based of the same code.

farhatahmad:
@defnull is correct. If this is still something you need, please create a PR matching there servers:yaml format (servers:json)

PR #258, opened 2020-06-25
Update scalelite_batch_import.sh

redesufpel:
If you try to run this script manually, the error message will print "sudo: ./post_publish/scalelite_post_publish.rb: command not found"
Just added 'ruby' before ./post_publish/.. to run correctly.
see ya


jgribonvald:
This fixed the problem for me too !

PR #261, opened 2020-06-27
Scan only BBB servers which are online and enabled in scalelite 

ironiemix:
We had an issue that our monitoring (which relies on ``rake status``) failed after we disabled a lot (~300) of our BBB servers in scalelite without deleting them.  ``rake status`` took a long time because every offline server produced a small timeout delay, and because of the large number they added up to a reasonable duration.

This can easily be prevented, if servers which are disabled or offline in scalelite are skipped (at least servers that are known to be offline should be skipped - by skipping disabled servers your stats are getting of a little bit, too)


lib/tasks/status.rake
```diff
@@ -9,21 +9,26 @@ task status: :environment do
...
       video_streams = 0
       users_in_largest_meeting = 0
 
-      meetings.each do |meeting|
-        count = meeting.at_xpath('participantCount')
-        users = count.present? ? count.text.to_i : 0
-        server_users += users
-        users_in_largest_meeting = users if users > users_in_largest_meeting
+      # Scan only enabled servers that are online
+      if server.online && server.enabled
```
mabeyj, line 19: +1, but it would be nice if this continues fetching the status of online disabled servers:

```suggestion
      # Scan only online servers
      if server.online
```

For maintenance, we disable the server and then keep checking the status before proceeding until it says no meetings so that we don't disrupt any meetings. If only enabled servers are fetched, it makes it harder to determine if any disabled servers are currently in use.

ironiemix: Yes, that is right, I did miss that!

agru:
Hi, it would be great if it would be possible to merge this to the next release. After updates we still have this problem.

lib/tasks/status.rake
farhatahmad: Yeah I would agree here. If you can change this to only check online servers, this can be merged

mschwrdtnr:
@jfederico Do I see it right, that you merged this without the suggestion of @mabeyj and @farhatahmad and disabled servers are not fetched anymore? This is necessary for our needs.

PR #288, opened 2020-07-26
URL matching for control commands

lkiesow:
This patch allows all commands to accept not only an exact ID match but
also search patterns for matching URLs. This allows for a much easier
control of groups of servers instead of having to find out the servers
identifier every time and then handle each server indivifually.

For example, you can run the following command to enbale multiple
servers:

    ./bin/rake 'servers:enable[bbb0\[1-5\].lkiesow.io]'

You can still use exact identifier matches though and the identifiers
will always be checked before checking the URL. Hence, you still can use
a command like:

    ./bin/rake 'servers:disable[a18201ec-f090-4bff-aaac-d0f5ab003f42]'

This closes #205 which was trying to solve the same problem by using the
host manes as server identifier instead. That shouldn't be necessary any
longer when this patch is applied.

---

This patch pretty much implements what we discussed in the community meeting a few weeks ago.


jgribonvald:
A such feature could be usefull when you are managing several BBB servers in backend.

lkiesow:
Updated this to be based on the current master branch

crosscodr:
I would love to have this feature. Will it be included in the next release?

jgribonvald:
> I would love to have this feature. Will it be included in the next release?

This is already available in the Last release, just need to set thé good property to be able to use it.

crosscodr:
> > I would love to have this feature. Will it be included in the next release?
> 
> This is already available in the Last release, just need to set thé good property to be able to use it.

Do you mean SERVER_ID_IS_HOSTNAME? To use this setting, I would be forced to re-add all servers to scalelite. I find this PR is better than that.

lkiesow:
@jgribonvald, as I wrote the original pull request #205 and this one: What was merged as #470 was a slightly modified version of my original pull request. Back then, jfederico had some complaints causing me to rework the patch into this pull request which lets you do more complex matching while still keeping the original identifiers for compatibility. I'm not sure about the reasoning behind reworking the old one and not merging this one, but the functionality is different.

jgribonvald:
I never said that your change aren't good ones, I just said that the change with the `SERVER_ID_IS_HOSTNAME` permit to use servernames, and yes you need to re-insert all BBB servers, but it's not a big effort...

crosscodr:
I think the main benefit from this PR is not just to use server-names as IDs, but to be able to match multiple server at once with a single command. I don't see the other PR provides this ability. Or am I wrong?

Issue #181, opened 2020-04-29
API to manage servers


rriscado:
Can we manage scalelite by API?
It's good for Add, Remove, Enable, Disable and Panc a Server with automations.

amassaro:
+1

congthang1:
API quite simple we just need to install a simple server (like Nodejs) to run list server commands and send responses. Hope this will be integrated soon.

farhatahmad:
Released in 1.5.1

PR #289, opened 2020-08-02
[#181] API to manage servers

aakatev:
Addresses #181 

---

The current PR adds an API to administrate the pool of BBB servers managed by Scalelite. The purpose is to mimic CLI commands using REST (mainly for the ease of configuration and management automation).

The following endpoints were added:

| Endpoint                        | Description                                  | Required Parameters      | Optional Parameters - |
| ------------------------------- | -------------------------------------------- | ------------------------ | --------------------- |
| scalelite/api                   | generic liveness probe endpoint                    | -                        | -                     |
| scalelite/api/getServers        | get all servers in the pool                  | -                        | -                     |
| scalelite/api/getServerInfo     | get info about particular server in the pool | serverID                 | -                     |
| scalelite/api/addServer         | add server in the pool                       | serverUrl, serverSecret  | loadMultiplier        |
| scalelite/api/removeServer      | remove server from the pool                  | serverID                 | -                     |
| scalelite/api/enableServer      | enable a server                              | serverID                 | -                     |
| scalelite/api/disableServer     | disable a server                             | serverID                 | -                     |
| scalelite/api/setLoadMultiplier | set load multiplier for a server             | serverID, loadMultiplier | -                     |

The endpoints are protected using [the same checksum scheme](https://docs.bigbluebutton.org/dev/api.html#api-calls) as BBB API. 

---

Two things:
- I chose not to expose an individual server secret through the endpoints for secuirty concerns (which i guess is arguable).
- I am not Rails developer and couldn't figure out how to run tests, therefore I didn't write test for the scalelite controller. If you are interested to merge the PR, I can work on tests more.

---

I also created a JavaScript wrapper library for the API and added some [tests there](https://github.com/aakatev/scalelite-js/blob/master/test/index.js).


umutcetinkaya:
It is absolutely beautiful. When would you add?

andrekutianski:
this is a must-have, thanks @aakatev

aakatev:
> this is a must-have, thanks @aakatev

Sure thing. I wouldn't expect this to be merged though. However, I keep the fork synced with upstream, so it should be pretty straightforward to start using this api in your setup, especially considering modular design of the scalelite.

aakatev:
If anyone still interested in using the API, I moved the controller into separate repo and it can be run as a sidecar for your scalelite. 

https://github.com/aakatev/scalelite-api

fireba11:
Is there any specific reason *not* to implement at least the first 3 (e.g. readonly) api paths?
That would solve #631 ... currently monitoring scalelite is a major PITA since i'd have to do docker exec rake commands on the the shell, so my monitoring would need specific sudo rights or similar. or i'd need a cronjob writing the current status to a temp file where my icinga plugin can try to parse it ... all very error-prone and unelegant.

Issue #346, opened 2020-10-21
Full disk with heavy load in production


jfederico:
When having scalelite in production with heavy load for a few days the logs grow as crazy. Sysadmins ether have to clean the logs manually or patch the systemd scripts

/var/lib/docker/containers/xxxx/xxxx.log

Same as for the docker images, after frequent updates, they tend to overload the disk

Solution

- Rotate logs on docker
- Prune images after update

PR #347, opened 2020-10-21
#346 GIT-346: updates to systemd script for rotating logs and prune docker…

jfederico:
… images


farhatahmad:
Should we be force image pruning? Some users may have multiple images that they switch between, and unless they always have a container attached to that image, the prune would delete it. This would probably cause some issues for users who might think that their images are dissapearing

jfederico:
Yeah, it is not critical. But we are not deleting the image from DockerHub, and we are doing pull on start

>ExecStartPre=/usr/bin/docker pull blindsidenetwks/scalelite:${SCALELITE_TAG}-api

Prune is run Post start, meaning that the a container is already attached.

What would be the point of keeping multiple images? In the end if I want to go back in versions by editing SCALELITE_TAG it will be pulled again

We had the issue that when some scalelite servers were automatically updated (by restarting scalelite.target) the new images were pulled, but the storage was clogged with non used images.

PR #371, opened 2020-11-17
Allow immediate recording cleanup

lkiesow:
This patch adds an option to the post publish script allowing for an
immediate deletion of the local recording after it has been successfully
transferred to the spool directory.

The default of the new option is false, keeping the current behavior.

This patch requires [BigBlueButton 2.2.30](https://github.com/bigbluebutton/bigbluebutton/pull/10841)


lkiesow:
Fixed style issue.

lkiesow:
Since [2.2.30](https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.2.30) has been released, this should now work with no additional adjustments

PR #440, opened 2021-02-17
SL-83: Make timeouts on requests to BBB api configurable

### LINKED ISSUE #439: Long timeout on requests to down servers causes response time issues
Right now when an API call is made that gets sent to a down server (e.g. getMeetingInfo, create on an existing meeting), the API call will likely hit the timeout, which by default is 10 seconds.

It turns out that this timeout is long enough to tie up threads in both scalelite and other applications calling scalelite, creating response time issues that affect meetings on working servers as well.

To mitigate this issue, we plan to make the following changes:

* Adjust the default timeout to a shorter value
* When forwarding API calls to a server that has been marked as down by the poller, reduce the timeout even further.
* Add environment variables to allow tuning the timeouts per site

kepstin:
This fixes #439.

Simply making the timeouts shorter and allowing them to be configured to match the environment of a particular deployment should reduce the impact of attempting to make calls on unreachable/slow BigBlueButton servers.

In addition, the timeouts on servers that the poller has marked as offline are shortened, to avoid a pile-up of hanging threads on servers that we suspect aren't going to be able to respond.

### DIFF
--- a/app/controllers/concerns/api_helper.rb
+++ b/app/controllers/concerns/api_helper.rb
@@ -6,7 +6,6 @@ module ApiHelper
   extend ActiveSupport::Concern
   include BBBErrors
 
-  REQUEST_TIMEOUT = 10
   CHECKSUM_LENGTH = 40
 
   # Verify checksum
@@ -39,7 +38,7 @@ module ApiHelper
   end
 
   # GET/POST request
-  def get_post_req(uri, body = '')
+  def get_post_req(uri, body = '', **options)
     # If body is passed and has a value, setup POST request
     if body.present?
       req = Net::HTTP::Post.new(uri.request_uri)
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/api_helper.rb
```diff
@@ -6,7 +6,7 @@ module ApiHelper
   extend ActiveSupport::Concern
   include BBBErrors
 
-  REQUEST_TIMEOUT = 10
+  DEFAULT_REQUEST_TIMEOUT = 5
```
jfederico, line 9: This is no longer used? It can be removed

jfederico (changes requested):
Should `response = get_post_req(uri, body, **bbb_req_timeout(server))` apply also to get_meetings and rake tasks?

kepstin:
The `getMeetings` api doesn't need the dynamic timeout code, because it already completely skips offline servers.

There doesn't seem to be a need for it in the rake tasks - there's no need to worry about blocking threads causing response time issues, and we want to do a full check on servers in the poller. A custom timeout can manually be set for the poller if needed using the environment variables.

jfederico (approved):

PR #442, opened 2021-02-17
SL:71 Override parameters coming from 3rd party

amalrails:
There are some settings that can be sent to BBB on create or join that will change the ones  predefined in BBB and can modify the features or behaviour of the HTML5 Client. This may override the parameters that are a hard requirements in some deployments.


app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -21,13 +21,12 @@ def index
   end
 
   def get_meeting_info
```
kepstin, line 23: Please remove these unrelated changes; there's no reason for this PR to touch API methods other than create and join.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -137,17 +135,16 @@ def get_meetings_disabled
   end
 
   def create
-    params.require(:meetingID)
-
+    meeting_id = require_meeting_id
```
kepstin, line 138: Creating the new function `require_meeting_id` is redundant, since it just runs `params.require(:meetingID)` the same as the code it's replacing!

If you really want `meeting_id` in a separate variable, then you could have done
```ruby
meeting_id = params.require(:meetingID)
```

That said, this change is also unrelated to the topic of the PR.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -164,9 +161,11 @@ def create
       params[:duration] = Rails.configuration.x.max_meeting_duration
     end
 
-    logger.debug("Creating meeting #{params[:meetingID]} on BigBlueButton server #{server.id}")
-    # Pass along all params except the built in rails ones
-    uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params)
+    logger.debug("Creating meeting #{meeting_id} on BigBlueButton server #{server.id}")
+    callee_method = __callee__.to_s.upcase
```
kepstin, line 165: This is a bit of a mess, please don't do this metaprogramming thing with method names.

You should either have different methods to check parameters for create vs. join (you can still factor out any shared code), or have the controller method pass the list of parameters to block rather than its own name.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -187,21 +186,20 @@ def create
   end
 
   def end
```
kepstin, line 188: Unrelated changes to `end` api shouldn't be in this PR.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -251,8 +249,8 @@ def join
 
   def get_recordings
```
kepstin, line 250: Unrelated changes to recording apis shouldn't be in this PR.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -379,8 +377,14 @@ def recordings_disabled
...
   # Filter out unneeded params when passing through to join and create calls
   # Has to be to_unsafe_hash since to_h only accepts permitted attributes
-  def pass_through_params
-    params.except(:format, :controller, :action, :checksum).to_unsafe_hash
+
+  def pass_through_params(name)
+    if ENV["#{name}_EXCLUDE_PARAMS"].nil?
+      params.except(:format, :controller, :action, :checksum).to_unsafe_hash
+    else
+      params.except(*(ENV["#{name}_EXCLUDE_PARAMS"].split(',').map(&:strip) +
```
kepstin, line 385: Rather than parse the environment variables at runtime on every request, please parse them to two separate config variables - e.g. `config.x.create_exclude_params` and `config.x.join_exclude_params` in `config/application.rb`.

test/controllers/bigbluebutton_api_controller_test.rb
```diff
@@ -537,6 +537,21 @@ class BigBlueButtonApiControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to encode_bbb_uri('join', server1.url, server1.secret, params).to_s
   end
 
+  test 'redirects user to the corrent join url with only permitted params for join' do
+    server1 = Server.create(url: 'https://test-1.example.com/bigbluebutton/api/',
+                            secret: 'test-1-secret', enabled: true, load: 0)
+    meeting = Meeting.find_or_create_with_server('test-meeting-1', server1)
+    params = { meetingID: meeting.id, password: 'test-password', fullName: 'test-name', test1: '', test2: '' }
+
+    mock_env('JOIN_EXCLUDE_PARAMS' => 'test1, test2') do
```
kepstin, line 546: Hmm. The one problem with doing the parsing once at startup is that it means changing the env at runtime might not work.

If we can't get this to work, another option would be to staticly configure some excluded params in the `.env.test` file, just make sure they don't conflict with any used in other tests.

kepstin (changes requested):

app/controllers/bigbluebutton_api_controller.rb
amalrails: it seemed better and more readable to move params.require(:meetingID) to a different method and just calling meeting_id wherever required instead of params[:meeting_id]

config/application.rb
```diff
@@ -85,5 +85,11 @@ class Application < Rails::Application
     config.x.recording_unpublish_dir = File.absolute_path(
       ENV.fetch('RECORDING_UNPUBLISH_DIR') { '/var/bigbluebutton/unpublished' }
     )
+
+    # List of BBB server attributes that should not be modified by create API call
+    config.x.create_exclude_params = ENV['CREATE_EXCLUDE_PARAMS']
```
kepstin, line 90: Please parse the values into an array here rather than later when the value is used. For example:
```suggestion
    config.x.create_exclude_params = ENV['CREATE_EXCLUDE_PARAMS']&.split(',') || []
```
which will set it to either a list of params or if the environment variable is nil, it'll set an empty list. Same thing for the `join_exclude_params` below.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -380,7 +380,12 @@ def recordings_disabled
   # Filter out unneeded params when passing through to join and create calls
   # Has to be to_unsafe_hash since to_h only accepts permitted attributes
   def pass_through_params
-    params.except(:format, :controller, :action, :checksum).to_unsafe_hash
+    # Get list of params that should not be modified by create/join API calls
+    excluded_params = Rails.configuration.x.send("#{action_name}_exclude_params")
```
kepstin, line 384: Rather than look up the configuration from the action name, please have the controller action pass the correct excluded parameters list as an argument to the `pass_through_params` method.

.env.test
```diff
@@ -0,0 +1,2 @@
+CREATE_EXCLUDE_PARAMS=""
+JOIN_EXCLUDE_PARAMS="test1, test2"
```
kepstin, line 2: The exclude params list should not contain a space - it's comma-separated, not "comma + space"-separated.

test/controllers/bigbluebutton_api_controller_test.rb
```diff
@@ -653,6 +653,19 @@ class BigBlueButtonApiControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to encode_bbb_uri('join', server1.url, server1.secret, params).to_s
   end
 
+  test 'redirects user to the corrent join url with only permitted params for join' do
```
kepstin, line 656: For testing the controller methods, maybe you can stub the `Rails.configuration.x.create_exclude_params` method? That would solve the issue of being able to test multiple different settings at runtime.

You still also need a separate test to make sure that the parsing in `config/application/rb` produced the expected values.

kepstin (changes requested):

.env.test
kepstin: (This also needs to be updated in the README)

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -166,7 +166,7 @@ def create
 
     logger.debug("Creating meeting #{params[:meetingID]} on BigBlueButton server #{server.id}")
     # Pass along all params except the built in rails ones
-    uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params)
+    uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params('create'))
```
kepstin, line 169: Pass the list itself, not the name.
```suggestion
    uri = encode_bbb_uri(
      'create',
      server.url,
      server.secret,
      pass_through_params(Rails.configuration.x.create_exclude_params)
    )
```

kepstin (changes requested):

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -165,8 +165,12 @@ def create
     end
 
     logger.debug("Creating meeting #{params[:meetingID]} on BigBlueButton server #{server.id}")
-    # Pass along all params except the built in rails ones
-    uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params)
+
+    # Get list of params that should not be modified by create API call
+    excluded_params = Rails.configuration.x.send('create_exclude_params')
```
kepstin, line 170: You don't need to use the `send` method to make a static method call, just do `Rails.configuration.x.create_exclude_params`

kepstin (changes requested):

kepstin (commented):
The code looks ready to merge now, but I'd really like to see more tests - cover both the 'join' and 'create' calls, and make sure it works with both an empty list and with some exclude parameters set.

kepstin (approved):

jfederico (changes requested):
After the last updates were merged this PR is not passing the CI. @amalrails can you please update them?

jfederico (approved):

PR #445, opened 2021-02-23
SL:70 Change logic for calculating server load

amalrails:
Change the algorithm for calculating the load on every server: 
New Logic: Server load calculation is now based on number of users, and the SL gives every meeting an assignment of 15 users when it starts, and then post that uses max(actual number of users, 15).

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,8 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `POLLER_THREADS`: The number of threads to run in the poller process. The default is 5.
 * `CONNECT_TIMEOUT`: The timeout for establishing a network connection to the BigBlueButton server in the load balancer and poller in seconds. Default is 5 seconds. Floating point numbers can be used for timeouts less than 1 second.
 * `RESPONSE_TIMEOUT`: The timeout to wait for a response after sending a request to the BigBlueButton server in the load balancer and poller in seconds. Default is 10 seconds. Floating point numbers can be used for timeouts less than 1 second.
+* `LOAD_MIN_USER_COUNT`: Minimum user count of a meeting, used for calculating server load. Defaults to 15.
+* `LOAD_JOIN_BUFFER_TIME`: The time(in minutes) until the `LOAD_MIN_USER_COUNT` will be used for calculating server load. Defaults to 15.
 
 ### Redis Connection (`config/redis_store.yml`)
 
--- a/lib/tasks/poll.rake
+++ b/lib/tasks/poll.rake
@@ -38,19 +38,32 @@ namespace :poll do
         resp = get_post_req(encode_bbb_uri('getMeetings', server.url, server.secret))
         meetings = resp.xpath('/response/meetings/meeting')
 
+        total_attendees = 0
+        load_min_user_count = Rails.configuration.x.load_min_user_count
+        x_minutes_ago = Rails.configuration.x.load_join_buffer_time.ago
+
+        meetings.each do |meeting|
+          created_time = Time.zone.at(meeting.xpath('.//createTime').text.to_i / 1000)
+          actual_attendees = meeting.xpath('.//participantCount').text.to_i + meeting.xpath('.//moderatorCount').text.to_i
+          total_attendees += if created_time > x_minutes_ago
+                               [actual_attendees, load_min_user_count].max
+                             else
+                               actual_attendees
+                             end
+        end
         # Reset unhealthy counter so that only consecutive unhealthy calls are counted
         server.reset_unhealthy_counter
 
         if server.online
           # Update the load if the server is currently online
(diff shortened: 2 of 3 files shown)


pielonet:
Hi, wouldn't it be even wiser to use this metric :
Sum up for each room the product : (number of users in the room * (number of active webcams + 1 (if screen sharing))
It would be roughly the number of video flows treated by the server, which is IMHO what really "loads" the server.

README.md
```diff
@@ -148,6 +148,8 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `POLLER_THREADS`: The number of threads to run in the poller process. The default is 5.
 * `CONNECT_TIMEOUT`: The timeout for establishing a network connection to the BigBlueButton server in the load balancer and poller in seconds. Default is 5 seconds. Floating point numbers can be used for timeouts less than 1 second.
 * `RESPONSE_TIMEOUT`: The timeout to wait for a response after sending a request to the BigBlueButton server in the load balancer and poller in seconds. Default is 10 seconds. Floating point numbers can be used for timeouts less than 1 second.
+* `MEETINGS_MIN_USER_COUNT`: Minimum user count of a meeting, used for calculating server load.
+* `MEETING_JOIN_BUFFER_TIME`: The time until the `MEETINGS_MIN_USER_COUNT` will be used for calculating server load.
```
kepstin, line 152: Please prefix these variables with `LOAD_` instead of `MEETING_` so that people don't confuse them with limits on the actual live meetings.

Make sure the documentation specifies the units for the `JOIN_BUFFER_TIME` (minutes)

lib/tasks/poll.rake
```diff
@@ -38,19 +38,32 @@ namespace :poll do
         resp = get_post_req(encode_bbb_uri('getMeetings', server.url, server.secret))
         meetings = resp.xpath('/response/meetings/meeting')
 
+        total_attendees = 0
+        meetings_min_user_count = ENV['MEETINGS_MIN_USER_COUNT'] || 15
+        x_minutes_ago = (ENV['MEETING_JOIN_BUFFER_TIME'] || 15).minutes.ago
```
kepstin, line 43: Environment variables are string type, so you get some type errors here when the environment variables are set.

I recommend reading the environment variables in `config/application.rb` and handle parsing them to integers and default values there, like
```ruby
config.x.load_min_user_count = ENV.fetch('LOAD_MIN_USER_COUNT', 15).to_i
```
then you can do
```ruby
x_minutes_ago = Rails.configuration.x.load_join_buffer_time.ago
```

kepstin (changes requested):

README.md
```diff
@@ -148,6 +148,8 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `POLLER_THREADS`: The number of threads to run in the poller process. The default is 5.
 * `CONNECT_TIMEOUT`: The timeout for establishing a network connection to the BigBlueButton server in the load balancer and poller in seconds. Default is 5 seconds. Floating point numbers can be used for timeouts less than 1 second.
 * `RESPONSE_TIMEOUT`: The timeout to wait for a response after sending a request to the BigBlueButton server in the load balancer and poller in seconds. Default is 10 seconds. Floating point numbers can be used for timeouts less than 1 second.
+* `LOAD_MIN_USER_COUNT`: Minimum user count of a meeting, used for calculating server load. Defaults to 15.
+* `LOAD_JOIN_BUFFER_TIME`: The time(in minutes) until the `LOAD_MIN_USER_COUNT` will be used for calculating server load. Defaults to 15.
```
kepstin, line 152: Hmm. The wording here is confusing or backwards - it implies that LOAD_MIN_USER_COUNT is not applied for the first 15 minutes, but then is applied afterwards.

I'm not sure the best way to word these help messages would be, maybe it would help to reverse the order, something like this?
```suggestion
* `LOAD_JOIN_BUFFER_TIME`: During the buffer time after a meeting starts, the server load calculation accounts a boosted load value to the meeting to compensate for people who have not yet joined. Values is in minutes, defaults to 15.
* `LOAD_MIN_USER_COUNT`: The minimum number of people to assume a meeting will have during the join buffer time. Defaults to 15.
```

kepstin (changes requested):

kepstin:
Code looks ready to merge, I just want to make sure the documentation is clear so that users will understand how the new environment variables work.

README.md
amalrails: hmm, may be @jfederico can help

christmart:
If I read the pull request correctly you don't use the load_multiplier anymore.
But it is still needed to distinguish different capable servers.

README.md
mschwrdtnr: @kepstin I would agree with you. I didn't get it with the readme explanation and had the need to look into the code.

PR #446, opened 2021-02-23
Compute load based on video streams

pielonet:
Hi,

I propose this pull request based on this one https://github.com/blindsidenetworks/scalelite/pull/445
I add the enhancement that load is actually computed based on actual audio and video streams.
We sum up 3 values :
* Normal audio users count for 1
* Webcam users count for 10 (a webcam stream is roughly equivalent to 10 audio streams)
* A screenshare counts for 20 (a sharescreen stream is roughly equivalent to 2 webcam streams)
These values are multiplied by the actual number of participants to get the actual number of streams treated by the server (SFU principle).


einhirn:
Perhaps this can be combined with #476 - this uses a configurable base load for the whole time (just reverts to actual user count after a configurable time), the PR I created with @defnull's code computes the actual count of video and audio streams.

PR #428, opened 2021-02-09
SL-65 & SL-66: Run scalelite poller parallely and Make poller timeout configurable

amalrails:
Test run with 4 bbb servers, polling interval 0 for 10000 times

with concurrency

real	3m36.787s
user	1m47.659s
sys	0m20.469s

without concurrency

real	11m2.587s
user	2m2.871s
sys	0m19.233s

PR #452, opened 2021-02-26
renamed to documented POLL_INTERVAL and default

timm2k:
Documentation talks about `POLL_INTERVAL` and there is no default value given.
Don't know if the syntax is correct. I'm not a ruby guy.


timm2k:
#428

kepstin:
The documentation actually refers to a variable in the [startup script for the docker image](https://github.com/blindsidenetworks/scalelite/blob/master/bin/start-poller), the change to the default poll interval was included in #428 by mistake.

I'm reverting the change in #454 to restore the previous behaviour.

Issue #152, opened 2020-04-12
Same server can be added multiple times


robbi5:
While trying to automate our bbb- and scalelite setup, I've found out that the servers model doesn't seem to use an unique constraint for URL and Secret, so that the same server got added multiple times:

```
$ ./bin/rake servers:add[https://bbb01.example.org/bigbluebutton/api,SECRETSECRETSECRET]
OK
id: ddf9c234-5bae-4b32-9ebf-11848d13d6b2
$ ./bin/rake servers:add[https://bbb01.example.org/bigbluebutton/api,SECRETSECRETSECRET]
OK
id: ba21e2f4-1fa1-4706-a09c-a71d7d23a129
$ ./bin/rake servers
id: ddf9c234-5bae-4b32-9ebf-11848d13d6b2
	url: https://bbb01.example.org/bigbluebutton/api
	secret: SECRETSECRETSECRET
	disabled
	load: unavailable
	offline
id: ba21e2f4-1fa1-4706-a09c-a71d7d23a129
	url: https://bbb01.example.org/bigbluebutton/api
	secret: SECRETSECRETSECRET
	disabled
	load: unavailable
	offline
```

I'm not sure if this is known and on purpose - if so, just close this issue :) But if not, maybe introducing an unique constraint for URL + Secret looks like a good idea to me.

farhatahmad:
Hi @robbi5 

Thanks for reporting this. This probably shouldn't be happening so I'll tag it as a bug

kepstin:
As part of the original design, I didn't think this was important... An administrator should just not add duplicate servers if they don't want them. (I can't think of any reasonable case to have them, but it doesn't break anything in scalelite if you do have them).

That said, we have a potential solution to this problem in #205 - by making the hostname itself the id rather than generating a unique uuid, it's much easier to ensure uniqueness.

jfederico:
This was solved in v1.0.12 and it is part of v1.1

PR #456, opened 2021-02-26
SL:87 validate uniqueness of server

amalrails:
Issue link [#152 ](https://github.com/blindsidenetworks/scalelite/issues/152)


app/models/server.rb
```diff
@@ -143,6 +143,18 @@ def reset_unhealthy_counter
...
 
+  # Validates uniqueness of a new server object, and attempts to save it. Raises a RecordNotSaved exception on errors,
+  # otherwise returns the newly created object.
+  def self.create!(attributes = {})
+    Server.all.each_slice(1000) do |server_list|
+      server_list.each do |server|
+        raise RecordNotSaved.new('Server url is not unique', self) if server.url.eql?(attributes[:url])
+        raise RecordNotSaved.new('Server id is not unique', self) if server.secret.eql?(attributes[:secret])
+      end
+    end
```
kepstin, line 154: There's no need to scan through all of the servers - you can simply check for the existence of the server id in the 'servers' set. That said, there's still a race condition in that case.

To do this correctly, the `create!` method needs to add the server inside a redis transaction (redis.multi) with a `WATCH` on the 'servers' set so that the transaction will abort if there's any concurrent modifications that add or remove servers.

kepstin (changes requested):

app/models/server.rb
kepstin: Oh, I'm getting this mixed up with unique server ids. Yeah, solving the issue of unique server urls (I wouldn't bother checking secret) is gonna be a pain to do, if we wanted to support it we should be adding an extra set to contain servers urls that can be checked against.

The original design included no protection against duplicate servers.

Note that checking only in create is not sufficient, because the url can be updated.

kepstin:
I don't think this should be merged, #205 is a better solution to this problem.

jfederico:
Fixed by #470

jfederico:
SL-140 Testing

Issue #179, opened 2020-04-27
Panicking a server fails because of missing moderator password.


defnull:
BBBs `end` API requires the moderator password, which the `panic` rake task does not provide. The meetings are removed from redis, but not actually ended.

https://github.com/blindsidenetworks/scalelite/blob/76105c117c051334277578799c098dfe1313cac4/lib/tasks/servers.rake#L67

This results in inconsistent state: Scalelite does not know about the meeting, will deny that it exists if asked directly (via `getMeetingInfo` or `isMeetingRunning`) but still includes it in a `getMeetings` call.

farhatahmad:
This was supposed to be fixed on the BigBlueButton side. If they don't change it, Scalelite will need to start storing moderatorPasswords that are passed on create

kepstin:
I guess we're not gonna get this fixed in the upstream BBB any time soon, so we're gonna have to start storing the moderator passwords.

PR #466, opened 2021-03-01
SL:88#179 add password attribute to meeting 

amalrails:
#179


kepstin (changes requested):
Using a static moderator password is not the way to handle this - if the moderator password is not passed on the request, a *random per-meeting* moderator password must be generated.

The moderator password should be saved per-meeting in the meeting hash in redis.

I recommend having scalelite save the moderator password used in the received create call, if there was none passed then generate a new random one, and then pass that password on to the backend BBB server.

The moderator password should be saved in redis in the meeting model.

app/models/meeting.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 class Meeting < ApplicationRedisRecord
-  define_attribute_methods :id, :server_id
+  define_attribute_methods :id, :server_id, :password
```
kepstin, line 4: Please call the attribute `moderator_pw` so we don't get it confused with the attendee password.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -147,7 +147,9 @@ def create
 
     # Create meeting in database
     logger.debug("Creating meeting #{params[:meetingID]} in database.")
-    meeting = Meeting.find_or_create_with_server(params[:meetingID], server)
+
+    moderator_pwd = params[:moderatorPW] || Rails.configuration.x.moderator_pwd
```
kepstin, line 151: Need to generate a new random password rather than using a static password.

kepstin (changes requested):

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -147,7 +147,10 @@ def create
 
     # Create meeting in database
     logger.debug("Creating meeting #{params[:meetingID]} in database.")
-    meeting = Meeting.find_or_create_with_server(params[:meetingID], server)
+
+    moderator_pwd = params[:moderatorPW] || SecureRandom.hex(6)
```
kepstin, line 151: We should probably try to match BBB itself as closely as possible in the random password generation: the relevant code is here: https://github.com/bigbluebutton/bigbluebutton/blob/363c80133895ca68c7825b463f8ca22df2d7f918/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L647 (It generates a random alphanumeric 8-character string for the password using this function: http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/RandomStringUtils.html#randomAlphanumeric-int- )

kepstin: It looks like the `SecureRandom` class has this function builtin, you could so `SecureRandom.alphanumeric(8)`

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -147,7 +147,10 @@ def create
 
     # Create meeting in database
     logger.debug("Creating meeting #{params[:meetingID]} in database.")
-    meeting = Meeting.find_or_create_with_server(params[:meetingID], server)
+
+    moderator_pwd = params[:moderatorPW] || SecureRandom.alphanumeric(8)
```
kepstin, line 151: We need to handle the case of a blank value (empty string) here. Rails provides a helper for this:
```suggestion
    moderator_pwd = params[:moderatorPW].presence || SecureRandom.alphanumeric(8)
```

kepstin (changes requested):

jfederico (approved):

kepstin (approved):

PR #470, opened 2021-03-01
Allow using server hostname as ID instead of UUID

### LINKED ISSUE #469: Allow using server hostname as ID instead of UUID
Opening an issue to track the feature request that was made in PR #205 

> When managing a larger cluster and wanting to take down specific hosts
> or chunks, is it quite inconvenient to find out the specific IDs handed
> out to servers in scalelite. Being able to use the domain names instead
> would be much nicer.

Some design requirements:
* This feature should be off by default, at least until a new "major" semver scalelite release, for compatibility reasons. (Enable via ENV variable.)
* When using `id` values based on user-provided data, a uniqueness check needs to be performed on the `id` field. (This is currently skipped because the UUIDs are assumed to be unique).

kepstin:
Closes #469 

This is an updated version of #205 that solves the problems that PR had:
* The hostname as id is now checked for uniqueness.
* The setting is off by default for compatibility reasons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `RESPONSE_TIMEOUT`: The timeout to wait for a response after sending a request to the BigBlueButton server in the load balancer and poller in seconds. Default is 10 seconds. Floating point numbers can be used for timeouts less than 1 second.
 * `LOAD_MIN_USER_COUNT`: Minimum user count of a meeting, used for calculating server load. Defaults to 15.
 * `LOAD_JOIN_BUFFER_TIME`: The time(in minutes) until the `LOAD_MIN_USER_COUNT` will be used for calculating server load. Defaults to 15.
+* `SERVER_ID_IS_HOSTNAME`: If set to "true", then instead of generating random UUIDs as the server ID when adding a server Scalelite will use the hostname of the server as the id. Server hostnames will be checked for uniqueness. Defaults to "false".
 
 ### Redis Connection (`config/redis_store.yml`)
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -150,6 +150,7 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `RESPONSE_TIMEOUT`: The timeout to wait for a response after sending a request to the BigBlueButton server in the load balancer and poller in seconds. Default is 10 seconds. Floating point numbers can be used for timeouts less than 1 second.
 * `LOAD_MIN_USER_COUNT`: Minimum user count of a meeting, used for calculating server load. Defaults to 15.
 * `LOAD_JOIN_BUFFER_TIME`: The time(in minutes) until the `LOAD_MIN_USER_COUNT` will be used for calculating server load. Defaults to 15.
+* `SERVER_ID_IS_HOSTNAME`: Instead of generating random UUIDs as the server ID when adding a server, use the hostname of the server as the id. Server hostnames will be checked for uniqueness.
```
amalrails, line 153: shouldn't we have a "defaulted  to `false`" here?

config/application.rb
```diff
@@ -92,6 +92,9 @@ class Application < Rails::Application
     # The time(in minutes) until the `load_min_user_count` will be used for calculating server load
     config.x.load_join_buffer_time = ENV.fetch('LOAD_JOIN_BUFFER_TIME', 15).to_i.minutes
 
+    # Whether to generate ids for servers based on the hostname rather than random UUIDs
+    config.x.server_id_is_hostname = ENV['SERVER_ID_IS_HOSTNAME']&.downcase == 'true'
```
jfederico, line 96: We should add a default value and in order to keep consistency at this point we should favour ENV.fetch

`config.x.recording_disabled = ENV.fetch('SERVER_ID_IS_HOSTNAME', 'false').casecmp?('true')`

Issue #471, opened 2021-03-02
1.0.10 Recording Importer Issue


semzor:
Importer doesn't import recordings from spool directory. Log related to this that I could find:

``Failed to import recording: undefined method `casecmp?' for nil:NilClass``

Do I need to explicitly set `RECORDING_DISABLED=false` in env file?

wolbernd:
Can confirm. Here is the full stacktrace:
```
W, [2021-03-01T11:08:50.490144 #6]  WARN -- : /srv/scalelite/lib/recording_importer.rb:12:in `import': undefined method `casecmp?' for nil:NilClass (NoMethodError)
        from /srv/scalelite/lib/tasks/recordings.rake:17:in `block (4 levels) in <main>'
        from /srv/scalelite/lib/tasks/recordings.rake:16:in `each'
        from /srv/scalelite/lib/tasks/recordings.rake:16:in `block (3 levels) in <main>'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/config.rb:23:in `call'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/processor.rb:115:in `_process_changes'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/processor.rb:19:in `block in loop_for'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/processor.rb:15:in `loop'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/processor.rb:15:in `loop_for'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/loop.rb:84:in `_wait_for_changes'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/event/loop.rb:42:in `block in setup'
        from /srv/scalelite/vendor/bundle/ruby/2.6.0/gems/listen-3.1.5/lib/listen/internals/thread_pool.rb:6:in `block in add'

wolbernd:
Small Update: The problem doesn't occur with v1.0.9 so going back to this version (at least for the recording importer) is a workaround.

jfederico:
[...]

PR #472, opened 2021-03-02
SL: 95, #471 Fix for undefined method `casecmp?' for nil:NilClass

amalrails:
#471

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -91,5 +91,8 @@ module Scalelite
 
     # The time(in minutes) until the `load_min_user_count` will be used for calculating server load
     config.x.load_join_buffer_time = ENV.fetch('LOAD_JOIN_BUFFER_TIME', 15).to_i.minutes
+
+    # Recording feature will be disabled, if set to 'true'. Defaults to false.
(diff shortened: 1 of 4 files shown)


config/application.rb
```diff
@@ -91,5 +91,8 @@ class Application < Rails::Application
 
     # The time(in minutes) until the `load_min_user_count` will be used for calculating server load
     config.x.load_join_buffer_time = ENV.fetch('LOAD_JOIN_BUFFER_TIME', 15).to_i.minutes
+
+    # Recording feature will be disabled, if set to 'true'. Defaults to false.
+    config.x.recording_disabled = ENV.fetch('RECORDING_DISABLED', 'false')
```
kepstin, line 96: If you do the conversion to a boolean here, e.g.
```ruby
config.x.recording_disabled = ENV.fetch('RECORDING_DISABLED', 'false).casecmp?('true')
```

then in the other checks below, you can just do conditionals like

```ruby
if Rails.configuration.x.recording_disabled
```

without needing further type conversions.

kepstin (changes requested):

kepstin (approved):

PR #476, opened 2021-03-04
make @defnull's proposition into a PR

einhirn:
I applied @defnull's code proposition [from #108](https://github.com/blindsidenetworks/scalelite/pull/108#issuecomment-621239375) to current master. Maybe this time the PR will be accepted.


einhirn:
Sorry - I don't understand rubocop's error message. Can someone with ruby knowledge help me fix this?

defnull:
We had this in production, but removed it again because it did not behave well in the situation described by follow-up comments. Also, my code sucks, I'm not a ruby developer.

einhirn:
> We had this in production, but removed it again because it did not behave well in the situation described by follow-up comments. Also, my code sucks, I'm not a ruby developer.

Combined with the newly added "assume configurable base load for the first few minutes"-feature it could work a bit better.

einhirn:
I guess this should also use
```
config.x.variable_name
```
instead of fetching Environment with every poll?

defnull:
> BTW: I don't know how to better attribute you, other than mentioning...

No worries, I don't care much about attribution here. I also signed a CLA so this PR being based on my work should not be a problem :)

defnull:
This looks wrong: `load += weight_users * total_attendees`. The total attendees number grows with each meeting, so you're counting users multiple times.

einhirn:
> This looks wrong: `load += weight_users * total_attendees`. The total attendees number grows with each meeting, so you're counting users multiple times.

Thanks for pointing that out - I'll adress that issue.

PR #459, opened 2021-02-28
SL-64: Scalelite should be compatible with BBB 2.3 recordings  closes #405, #416

jfederico:

PR #487, opened 2021-03-16
2.3 prototype

glxnet:
Update Dockerfile to use the bionic repository for the 2.3 version.
Add Makefile for dumping the created docker images (optional)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.11 AS alpine
 
-FROM ubuntu:16.04 AS bbb-playback
+FROM ubuntu:18.04 AS bbb-playback
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
     && apt-get install -y language-pack-en \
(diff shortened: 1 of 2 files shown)


jfederico:
Thanks @glxnet. But if that works for you I would suggest you to build your own image and use that.

The issue is not that simple. bbb 2.3 is still in active development and we can not endorse code for production that may change at any time. Building that image from bbb-23-dev would do exactly that. And this is precisely why this PR has not being merged https://github.com/blindsidenetworks/scalelite/pull/459

There is a different approach we are exploring that is use official BigBlueButton images for proxyng recordings see: https://github.com/bigbluebutton/bbb-playback-proxy and https://hub.docker.com/repository/docker/bigbluebutton/bbb-playback-proxy. Those images are built along the BBB packaging whenever the core developers merge or push update the main repo.

More changes to come to SL, such as decoupling the nginx front proxy from this packaging and make a one and single image for scalelite instead of 3. But that will likely be tagged as version 2, as it is a very disruptive change.

PR #493, opened 2021-03-17
SL-101: Re-include disabled servers in status rake task.

### LINKED ISSUE #481: [poller] Poller shows zero meetings immediately after disabling server
When running command 

> docker exec -it scalelite-api ./bin/rake status

number of meetings was shown for "disabled online" servers. Now, immediately after running
> docker exec -t scalelite-api ./bin/rake servers:disable[ID]

status shows 0 meetings 0 users, even if there is still meeting running on servers.

Screenshot from our zabbix, parsing results of ./bin/rake status, showing numbers users drop immediately after issuing servers:disable for bbb-nodes1-7:
2021-03-11-141726_606x566_scrot
docker inspect:
[poller_docker_inspect.txt](https://github.com/blindsidenetworks/scalelite/files/6123230/poller_docker_inspect.txt)

farhatahmad:
fixes #481


jfederico:
This was the first idea that crossed my mind (https://github.com/blindsidenetworks/scalelite/issues/481) and it was assigned to @amalrails . But I had second thoughts. Should we consider this approach only or also adding a new panic state or action for those servers that need to be put out of the pool but are still holding meetings?

jfederico (approved):

PR #501, opened 2021-03-23
Fix #500

### LINKED ISSUE #500: get_meetings issue when a server is manually disabled
When manually shutting down a server and then calling the "get_meetings" action.
After that scalelite will return an with an error 500.
This is because the code does not account for disabled servers.

thijs-sebsoft:
Using available will resolve potential server errors.


jfederico:
Thanks for the PR, but it is not passing the tests

jfederico:
Closing in favour of https://github.com/blindsidenetworks/scalelite/pull/506

PR #524, opened 2021-04-08
SL-97: add recording ready notification. Closes #485

### LINKED ISSUE #485: Question: Send notifications when a recording is ready?
BigBlueButton has ability to send notification when recording is ready. This is implemented in Moodle plugin.
How can we do this with Scalelite?

amalrails:
Implement core  post_publish scripts such as recording_ready_notification

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,7 @@ AllCops:
     - bin/update
     - bin/yarn
     - vendor/bundle/**/*
+    - db/migrate/*
   TargetRubyVersion: 2.6
 Layout/HashAlignment:
   EnforcedHashRocketStyle: [ key, table ]
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,7 +14,6 @@ AllCops:
     - bin/update
     - bin/yarn
     - vendor/bundle/**/*
-    - db/migrate/*
   TargetRubyVersion: 2.6
 Layout/HashAlignment:
   EnforcedHashRocketStyle: [ key, table ]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -76,6 +76,12 @@ GEM
     crack (0.4.5)
       rexml
     crass (1.0.6)
+    daemons (1.3.1)
+    delayed_job (4.1.9)
+      activesupport (>= 3.0, < 6.2)
+    delayed_job_active_record (4.1.6)
+      activerecord (>= 3.0, < 6.2)
+      delayed_job (>= 3.0, < 5)
     dotenv (2.7.6)
     dotenv-rails (2.7.6)
       dotenv (= 2.7.6)
@@ -94,6 +100,7 @@ GEM
     i18n (1.8.9)
       concurrent-ruby (~> 1.0)
     jaro_winkler (1.5.4)
+    jwt (2.2.2)
     listen (3.1.5)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
@@ -209,10 +216,13 @@ DEPENDENCIES
   bootsnap (>= 1.4.2)
   byebug
   connection_pool (~> 2.2.2)
+  daemons
+  delayed_job_active_record
   dotenv-rails
   factory_bot_rails
   fakeredis!
   hiredis (~> 0.6.3)
+  jwt
   listen (>= 3.0.5, < 3.2)
   minitest-stub_any_instance
   pg
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,7 +100,7 @@ GEM
     i18n (1.8.9)
       concurrent-ruby (~> 1.0)
     jaro_winkler (1.5.4)
-    jwt (2.2.2)
+    jwt (2.2.3)
     listen (3.1.5)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
@@ -222,7 +222,7 @@ DEPENDENCIES
   factory_bot_rails
   fakeredis!
   hiredis (~> 0.6.3)
-  jwt
+  jwt (= 2.2.3)
   listen (>= 3.0.5, < 3.2)
   minitest-stub_any_instance
   pg (~> 1.2)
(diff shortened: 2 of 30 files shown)


Gemfile.lock
```diff
@@ -209,10 +216,13 @@ DEPENDENCIES
   bootsnap (>= 1.4.2)
   byebug
   connection_pool (~> 2.2.2)
+  daemons
+  delayed_job_active_record
   dotenv-rails
   factory_bot_rails
   fakeredis!
   hiredis (~> 0.6.3)
+  jwt
```
jfederico, line 225: Lets freeze the version of this gem

README.md
```diff
@@ -155,6 +155,7 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `SERVER_ID_IS_HOSTNAME`: If set to "true", then instead of generating random UUIDs as the server ID when adding a server Scalelite will use the hostname of the server as the id. Server hostnames will be checked for uniqueness. Defaults to "false".
 * `CREATE_EXCLUDE_PARAMS`: List of BBB server attributes that should not be modified by create API call. Should be in the format 'CREATE_EXCLUDE_PARAMS="param1,param2,param3"'.
 * `JOIN_EXCLUDE_PARAMS`: List of BBB server attributes that should not be modified by join API call. Should be in the format 'JOIN_EXCLUDE_PARAMS="param1,param2,param3"'.
+* `RECORDING_READY_NOTIFIER_TIME`: Schedule recording ready notification to run after x minutes. Defaults to 2 minutes'.
```
jfederico, line 158: Why is this delay needed?, what happen if I set it up to 0?

app/jobs/recording_ready_notifier_job.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'jwt'
+require 'net/http'
+
+class RecordingReadyNotifierJob < ApplicationJob
+  queue_as :call_back
+
+  LOGGER = Delayed::Worker.logger
```
jfederico, line 9: All capitals?

.rubocop.yml
```diff
@@ -14,6 +14,7 @@ AllCops:
     - bin/update
     - bin/yarn
     - vendor/bundle/**/*
+    - db/migrate/*
```
kepstin, line 17: Migrations are normally written by hand and should follow the standard coding style, please don't exclude them all.

README.md
kepstin: I agree - no need for this setting to exist, just enqueue the notification to run immediately after the recording is published.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -174,8 +174,15 @@ def create
 
     # Pass along all params except the built in rails ones and excluded_params
     uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params(excluded_params))
+    callback_data = nil
 
     begin
+      if params['meta_bn-recording-ready-url']
```
kepstin, line 180: Use
```ruby
if params['meda_bn-recording-ready-url'].present?
```
to correctly handle the parameter being blank.

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -174,8 +174,15 @@ def create
...
+        callback_attributes = { recording_ready_url: params['meta_bn-recording-ready-url'] }
+        callback_data = CallbackData.create!(meeting_id: meeting.id, callback_attributes: callback_attributes)
```
kepstin, line 182: At this point you already have the `meeting` object, so you should be able to use:
```ruby
callback_data = meeting.callback_data.create!(callback_attributes: callback_attributes)
```

app/jobs/recording_ready_notifier_job.rb
kepstin: Inside a Rails ApplicationJob, you should be able to use the normal rails logger? Shouldn't be any need for this, I think…

(If there is a need to set up the logger specially for delayed job, it should be done in an initializer or in the delayed job startup script, not in the job itself)

config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,11 @@
...
+
+Delayed::Worker.destroy_failed_jobs = false
+Delayed::Worker.sleep_delay = 60
+Delayed::Worker.max_attempts = 3
+Delayed::Worker.max_run_time = 5.minutes
+Delayed::Worker.read_ahead = 10
+Delayed::Worker.default_queue_name = 'default'
+Delayed::Worker.delay_jobs = !Rails.env.test?
+Delayed::Worker.raise_signal_exceptions = :term
+Delayed::Worker.logger = Logger.new(Rails.root.join('log/delayed_job.log'))
```
kepstin, line 11: Unconditionally logging to a file here is problematic… This should use similar logic to the normal rails logging initializer. For use in docker, it should be logging to stdout/stderr?

test/models/callback_data_test.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
kepstin, line 1: An empty test file shouldn't be committed.

kepstin (changes requested):

app/controllers/bigbluebutton_api_controller.rb
amalrails: meeting is not an active record object so it wouldnt work

app/controllers/bigbluebutton_api_controller.rb
```diff
@@ -174,8 +174,15 @@ def create
 
     # Pass along all params except the built in rails ones and excluded_params
     uri = encode_bbb_uri('create', server.url, server.secret, pass_through_params(excluded_params))
+    callback_data = nil
 
     begin
+      if params['meta_bn-recording-ready-url'].present?
+        callback_attributes = { recording_ready_url: params['meta_bn-recording-ready-url'] }
```
jfederico, line 181: One detail here. I don't think it is documented but we need to make this backward compatible with the old meta parameters
See https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/post_publish/post_publish_recording_ready_callback.rb#L48-L51

jfederico (changes requested):

jfederico (approved):

PR #527, opened 2021-04-09
SL:117 Lock file in order to make sure there is only one recording-importer process at the time

amalrails:
Add a lock file in order to make sure there is only one recording-importer process at the time

### DIFF
--- a/lib/recording_importer.rb
+++ b/lib/recording_importer.rb
@@ -9,10 +9,14 @@ class RecordingImporter
   end
 
   def self.import(filename)
-    return if Rails.configuration.x.recording_disabled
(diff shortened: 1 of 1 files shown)


faust64:
Sounds nice, although running scalelite Docker images, `/tmp` would usually not be shared across containers. It could make sense to have this somewhere in /var/bigbluebutton, which is more likely to be shared.
Maybe even in /var/bigbluebutton/spool -- I'm curious to run multiple recordings importer, sharing the same /var/bigbluebutton, but with their own /var/bigbluebutton/spool, grouping BBBs nodes with their own recordings-importer, ... Maybe overkill, ...

PR #533, opened 2021-04-21
SL:110: add cordon state

amalrails:
New state for servers to retain existing meetings and not accept anymore meeting creation requests, ref: #500


jfederico:
This approach must be changed. We want 3 states [<disabled>|enabled|cordoned] not an extra flag

Besides, the implementation does not work. In order to cordon, the server is required to be disabled first and then cordoned, which does not make sense when putting to drain of a server for maintenance for example, you go from enabled to cordoned directly.

jfederico (approved):

PR #534, opened 2021-04-22
SL97: Implement core post_publish scripts such as recording_ready_notification

amalrails:


jfederico (approved):

jfederico:
@amalrails This implementation must be refactored in the next release so all the logic for handling extra parameters and validating them are taken out of the controller (which is bolted enough already) and the callbacks generalised.

Issue #545, opened 2021-05-10
Scalelite Recordings Metrics & BBB Exporter


faust64:
Hi,

Deploying Scalelite and the BigBlueButton Exporter ( https://github.com/greenstatic/bigbluebutton-exporter/ ), I noticed that while that exporter reports the right values for recordings (published / unpublished / update / deleted) when querying my BigBlueButton servers API, it seems to be broken when querying Scalelite.

The counters for published / unpublished / update / deleted recordings are all equals to the same value, which is the count of published recordings.

AFAIU, the exporter would query for some getRecordings API call, passing an argument ("deleted", "published", ...), over there:
https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/collector.py#L202

The recording state being sent as a param, in getRecording API command:
https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/api_lib.py#L86

Now, AFAIU, Scalelite implements 4 different  commands:
https://github.com/blindsidenetworks/scalelite/blob/5feb7cb3b5023c3b6243210659f665cc940f2b56/config/routes.rb#L23

Thus, when the exporter would query for getRecordings with different arguments, the same counter would be returned 4 times.

One way to deal with this might be to add some configuration option in the exporter, handling Scalelite specifics, ... although we could make the argument Scalelite should implement its API matching the way BigBlueButton works. Which would make more sense, IMHO.

Issue #193, opened 2020-05-06
getRecordings API not fully implemented 


defnull:
Scalelite currently does not implement the `state` and `meta` filter parameters for `getRecordings`. See https://github.com/bigbluebutton/greenlight/issues/1466 for why this might be required by a front-end application.

mabras:
Before Scalelite, we have used the `meta` to get the recording size, but not after Scalelite. It very helpful to get the `meta` parameters again.

defnull:
A `size` parameter is not is not part of the `getRecordings` API?

cemezgin:
Hi. Is there any solution for this problem? I could get the size data before scalelite implementation but now there is a blank array return for meta.

mehdikazemi3120:
Hi. I have the Same problem. please add **meta filter** parameters for getRecordings.
Tnx

cochennec:
Hi, we have the same problem here and would appreciate the "state" parameter to be implemented, please.

PR #546, opened 2021-05-10
fix(recordings): invalid metrics #545

faust64:
getRecordings API call should accept a `state` argument.

According to the BigBlueButton exporter, possible values include `published`, `unpublished`, `deleted` and `processing`:
https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/collector.py#L165..L207

While this patch doesn't check for a valid input, it should include the proper `WHERE` clause querying DB.

Tested against a setup I can't really play with, that only includes published records: counters are OK.
Additional tests & feedback welcome.

Fixes https://github.com/blindsidenetworks/scalelite/issues/545
Partly addresses https://github.com/blindsidenetworks/scalelite/issues/193


amalrails:
@faust64 thanks a lot for raising this pr, closing this pr with #575 which also includes test cases and more filters.

PR #575, opened 2021-06-02
SL-129 & SL-130 Add filter for get_recordings api

amalrails:
- Filter getRecordings request based on meta parameters

- Filter getRecordings request based on the recording state


amalrails:
fixes #545

jfederico (approved):

PR #578, opened 2021-06-08
add load column to rake status table

einhirn:
Hi, currently you can't see which server has which current load value - which could differ from the meetings and/or attendee count. This PR adds a "load" column to `rake status`


jfederico:
Hi @einhirn ,  thanks for the PR.

You need to keep the PR updated or give us access so we can update the branch before merging.

einhirn:
updated to "master"

jfederico (approved):

PR #583, opened 2021-06-10
SL-136 Add delete state to recordings

amalrails:


defnull:
Am I missing something here? How do _actually_ delete a recording now?

jfederico:
All the code was lost after the PR for fixing the state was merged and reverted. This PR needs to be re-opened.

PR #599, opened 2021-07-07
create #{meeting_id}-sender.done

ffdixon:


bigbluebutton/scalelite_delete_older_recordings.rb
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -x
+
+# Delete RAW recordings after 4 days
+MAXAGE=4
+
+# Delete events files (use for analytics data) after 2 weeks
+EVENTS_MAXAGE=14
+
+LOGFILE=/var/log/bigbluebutton/bbb-sender-cleanup.log
```
jfederico, line 9: Rubocop caught an error here because it is not enclosed in strings.

But, the new way of logging in BBB is sending them to stdout because they are now managed through journalctl. It also simplifies things for aggregation (although not for SLE).

In the other script we changed that to simply puts. For example:

puts "$(date --rfc-3339=seconds) Deleting sent recordings older than ${MAXAGE} days" 


If you want to log to a file, I would say we want to use the actual Logger instead. Such as in here:

https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/post_publish/post_publish_recording_ready_callback.rb#L29

For example

BigBlueButton.logger.info("$(date --rfc-3339=seconds) Deleting sent recordings older than ${MAXAGE} days" )

jfederico (changes requested):

bigbluebutton/scalelite_delete_older_recordings.rb
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -x
+
+# Delete RAW recordings after 4 days
+MAXAGE=4
```
jfederico, line 4: This should not be hardcoded, but based on ENV variables instead

jfederico (changes requested):

Issue #417, opened 2021-02-02
Remove recorded files older than N days in Scalelite


hmahdi-s:
Is there any script to remove older recorded files as cron job like [bbb server](https://docs.bigbluebutton.org/2.2/customize.html#delete-recordings-older-than-n-days)?

atikulbd:
Yes. Here is the url:
```
https://docs.bigbluebutton.org/2.2/customize.html#delete-recordings-older-than-n-days
```
Put the code inside:
```
/etc/cron.daily/bbb-recording-cleanup
```
And set permission to **755**
```
chmod +x /etc/cron.daily/bbb-recording-cleanup
```

You can change the value of ***MAXAGE*** to specify how many days of recordings you want to preserve.
Done

hmahdi-s:
@atikulbd 
I wrote that link. 
But i want to remove recorded files from scalelite server. not BBB server

atikulbd:
Ah! My mistake.
I will write a script for **Scalelite** and send you here shortly.

hmahdi-s:
@atikulbd 
That's kind of you
Thank you

amg-web:
here the script:
`#/bin/bash
URI=domain
SALT=SALT
MAX_AGE=60

MAX_DATE=$(date -I -d "-$MAX_AGE days")

for MEET in $(sudo -u postgres -H -- psql -d scalelite -c "SELECT record_id FROM recordings WHERE starttime < '$MAX_DATE';"); do

LINK="deleteRecordingsrecordID=$MEET$SALT"

KEY=$(echo -n "$LINK" | openssl sha1 |awk '{print $2}')

FLINK="https://$URI/bigbluebutton/api/deleteRecordings?recordID=$MEET&checksum=$KEY"
echo "deleting $MEET"
curl $FLINK
done`

faust64:
In your for look, I think you should also remove the recording from filesystem:

```
#/bin/bash

URI=domain
SALT=SALT

MAX_AGE=60
MAX_DATE=$(date -I -d "-$MAX_AGE days")

for MEET in $(sudo -u postgres -H -- psql -d scalelite -c "SELECT record_id FROM recordings WHERE starttime <'$MAX_DATE';")
do
    LINK="deleteRecordingsrecordID=$MEET$SALT"
    KEY=$(echo -n "$LINK" | openssl sha1 | awk '{print $2}')
[...]

PR #601, opened 2021-07-07
SL-186: Added logic to delete recordings sent to scalelite after a few days

ffdixon:
This pull request adds a cron.daily job to delete sent recordings from the BigBlueButton server after 4 days.


jfederico (approved):

jfederico:
Related to #417, we need to document it, but this pretty much should be added to a cron job in each BBB server

PR #607, opened 2021-07-13
SL-187:  Fix getRecordings() Redirect Issue

amalrails:
The issue is that going to the url of the recorded meeting from getRecordings() returns a 301 redirect that is non-https which violates the client's IT policy


jfederico (approved):

christmart:
The fix breaks the functionality if additional recording formats are registered and are not directories but files.
In our case we register for every recording a downloadable MP4 file. This fix adds a / to the end of the url which breaks the access.

The 301 redirect mentioned in the issue is generated by the webserver when accessing a directory. The redirect should be corrected in the webserver configuration to not return a http instead of https.

Please revert this change.

PR #614, opened 2021-07-25
SL-193: Recording multiple formats publish un-publish issue

amalrails:
Recording multiple formats publish un-publish issue


jfederico (approved):

defnull:
Migration skripts do not match schema.rb changes! Upgrading an existing scalelite DBs will result in a broken database. Please revert this PR before releasing.

PR #619, opened 2021-08-02
Fix db:migrate task and auto-bootstrap DB in bin/start

defnull:
Scalelite currently requires manual steps to bootstrap or migrate its database. Automating scalelite deployments proved to be difficult, specially in environments that follow basic security principles and do not grant CREATEDB privileges to application accounts. If an empty database already exists, then `db:setup` would fail and `db:migrate` unfortunately creates an incomplete schema. Calling `db:schema:load` on the other hand is destructive and should not be executed automatically. This leaves no real option for automated deployment.

This MR adds the missing database migration step and ensures that `db:migrate` on an existing but empty database works as expected. The `bin/start` script, which is used as a docker container entry point, is now able to automatically bootstrap or migrate the DB on startup (inspired by greenlight).


defnull:
I am not sure why rubocop complains this much over code that was literally copy&pasted from `schema.rb`, but does not complain about the same issues in `schema.rb`.

defnull:
The most recent beta broke migration again: 38dc3d40729dc0bab82f2b58c5ee311d83e9d669

Please be more careful with schema changes. Bad migration scripts mean that existing instances will silently break on update and admins have to throw away their database, causing downtime and confusion. If Scalelite does not support migration, then please document this better and do not bother adding incomplete migration scripts.

defnull:
If this PR is missing something, please advise. It is in review-state for a month now.

defnull:
The date was intentional because it matched the corresponding schema version. https://github.com/blindsidenetworks/scalelite/blob/6b175a4ec5b8c318d8c52256bc8590d9a82836a2/db/schema.rb#L13

defnull:
Please revert your change. The wrong date breaks migrations of old scalelite instances. Again.

To explain: Rails remembers which schema versions the DB is currently corresponds to. This is `20181123180008` for any DB created before 2021-04-25. Since you now created a migration step dated AFTER that, it will be applied, and that will fail because the tables already exist. Without your last-minute change, everything works as expected.

PR #624, opened 2021-08-03
add recording_not found template

amalrails:


app/controllers/playback_controller.rb
```diff
@@ -16,6 +19,8 @@ def play
       # If recording isn't protected, don't check tokens
       deliver_resource(true)
     end
+  rescue RecordingNotFoundError
```
kepstin, line 22: Rather than have the `rescue` inside this controller method, I think it would be better to  a `rescue_from` handler at the top controller level: see https://guides.rubyonrails.org/action_controller_overview.html#rescue-from

kepstin (changes requested):

kepstin (approved):

PR #645, opened 2021-08-24
Add rake task to sync cluster state (add/update/remove servers) from a YAML file

defnull:
For automated deployment, it would be extremely helpful to be able to get a scalelite cluster into a desired state based on a simple configuration file. The `servers:addAll` tasks goes into the right direction, but it is not able do modify or remove servers. The idea would be to have a rake task named `servers:sync[file,mode,dryrun]` that reads a YAML file and adds/removes/modifies servers as necessary. A second task called `servers:yaml[verbose]` can be used to bootstrap this configuration file from an existing scalelite cluster.

The YAML file would look like this:

```yaml
servers:
    <server-id>:                   # Must be unique, should be the server hostname
        secret: <string>           # required
        url: <string>              # default: "https://<server-id>/bigbluebutton/api"
        enabled: <bool>            # default: true
        load_multiplier: <float>   # default: 1.0, must be greater than 0

    # Example for a simple server (ID equals hostname)
    bbb1.example.com:
        secret: "1bdce5cbab581f3f20b199b970e53ae3c9d9df6392f79589bd58be020ed14535"

    # Example for a legacy server (generated ID) and some non-default settings
    02bff3a7-c95f-49d3-b1e5-c53eddd4dd68:
        secret: "2bdce5cbab581f3f20b199b970e53ae3c9d9df6392f79589bd58be020ed14535"
        url: "https://bbb2.example.com/bigbluebutton/api"
        enabled: false
        load_multiplier: 5.0
```

The `servers:sync[file,mode,dryrun]` task would read a YAML file and move the cluster to the desired state. More precisely, it would:

- [x] Add missing servers to the cluster.
- [x] Update config for existing servers (secret, url, load_multiplier) if necessary.
- [x] Cordon servers that are enabled but should be disabled.
- [x] Enable servers that are cordoned or disabled but should be enabled.
- [x] Remove servers from the cluster that are not defined in the YAML. This is affected by the `mode` parameter:
  - [x] In `keep` mode, no servers are removed.
  - [x] In `cordon` mode (default), empty servers are removed and servers with meetings on them are cordoned.
  - [x] In `force` mode, servers are removed after all meetings are forcefully ended (similar to `servers:panic`).
- [x] If a `dryrun` is requested, the task runs normally but does not persist any of the changes.

A second task called `servers:yaml[verbose]` can be used to bootstrap this configuration file from an existing scalelite cluster.

Feedback is welcomed (I'm not a Ruby developer). Whats missing?

- [x] Tests
- [x] Documentation
- [ ] Feedback from maintainers


defnull:
Added documentation and tests and finally got the implementation to work as expected. I have not tested the "remove non-empty server" case yet, so handle with care. I'd greatly appreciate some feedback at this point.

I hat do change `Server.save!` to allow custom IDs for new servers. The code there is a bit strange (e.g. the `SecureRandom.uuid` fallback is duplicated) but I tried to keep this PR small and forced myself to not refactor anything unrelated. I also want to say that I hate rubocop and ruby with a passion now.

defnull:
We now rely on this functionality (and #619) for automated deployments. I'll keep this PR alive if there is interest, but looks like we have to continue maintain a private fork anyway. So, no rush.

ffdixon:
Thanks for your work @defnull - we appreciate it (and your help with #619).  We've been focused on getting Scalelite 1.2 tested and released, which is imminent.  We'll circle back to this pull request after the 1.2 release.

defnull:
Good to know that there is some interest and a chance to get this merged after 1.2 release. I worked a bit more on this PR today and just force-pushed a couple of changes:

* There is now a `keep` mode that does not remove superfluous servers. The `panic` mode was renamed to `force`.
* There is now a `dryrun` feature to simulate a sync without actually changing anything.
* I also added a `servers:yaml` task to export the current cluster state in the same format that is accepted by `servers:sync`. This may also be useful for scripting, as there is currently no other way to get the current cluster state in an easily parseable standard format.

jfederico (approved):

PR #661, opened 2021-09-07
SL-201: Update README.md

amalrails:


jfederico:
When reading the main README page it is not clear what the `SCALELITE_TAG` is about, as it is only referenced in the docker-README.md file.  Shouldn't be this update note somewhere else?

There should be a card for the Systemd deployment itself with everything that is implied. Including preparing the env variables and setting up the files.

PR #670, opened 2021-09-13
SL- 209: Fix db:migrate task and auto-bootstrap DB in bin/start

amalrails:
This adds a missing DB migration step and ensures that 'db:migrate'
on an existing but empty DB works as expected. The bin/start script,
which is used as a docker container entry point, now bootstraps or
migrates the DB on startup (inspired by greenlight).

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -1,3 +1,13 @@
 #!/bin/sh
 
+db_create="$(bundle exec rake db:create 2>&1)"
+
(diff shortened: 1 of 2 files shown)


defnull:
Please double-check if you really want the auto-migrate-on-container-startup logic to be merged in a patch release.

jfederico (approved):

PR #682, opened 2021-09-27
SL-212: Create PROTECTED_RECORDINGS_README.md

amalrails:


jfederico (approved):

relecand:
Hello, thank you for the readme. What happens to old recordings? Would they have to be re-rendered or are they also protected with the new feature?

Issue #700, opened 2021-10-09
404 error on recordings playback + Old Recording not show after upgrade to v1.2.0.1


morexzamani:
Hi Scalelite development team (@ffdixon , @amalrails , @jfederico , @rabser )
I've faced a problem some days and unfortunately can't find anyone to solve it. Please help me.

I use Moodle to handle classes. My Scalelite server worked very well without any problem until I was updated it (If I`m not mistaken, from v1.1.7 to 1.2.0) one or two weeks ago. After that, my old recordings have gone from moodle recordings list and I have 404 error for new recordings that appear in the Moodle list after upgrade, 

For error 404 I found my recordings remain in the spool directory and scalelite-recording-importer is not working correctly 

(Note that I downgraded to v1.1.8.1, but the problem still exists!)

Please help me to return them back.

## Here is some reports:

**`nano /etc/default/scalelite`:**

```
URL_HOST=XXX.XXX.XXX
SECRET_KEY_BASE=XXXX
LOADBALANCER_SECRET=XXXXX
DATABASE_URL=postgresql://postgres:XXXXXXPASSWORD@XXXXIP:5432
REDIS_URL=redis://redis
NGINX_SSL=true
SCALELITE_NGINX_EXTRA_OPTS=--mount type=bind,source=/etc/letsencrypt,target=/etc/nginx/ssl,readonly
SCALELITE_TAG=v1.2.0.1
SCALELITE_RECORDING_DIR=/mnt/scalelite-recordings/var/bigbluebutton
POLLING_INTERVALL=2
```

**`docker ps`:**

```
CONTAINER ID   IMAGE                                                   COMMAND                  CREATED         STATUS         PORTS                                                                      NAMES
XXXX7338a945   blindsidenetwks/scalelite:v1.2.0.1-poller               "bin/start-poller"       6 minutes ago   Up 6 minutes                                                                              scalelite-poller
XXXX623904cd   blindsidenetwks/scalelite:v1.2.0.1-recording-importer   "bin/start-recording…"   6 minutes ago   Up 6 minutes                                                                              scalelite-recording-importer
XXXX24978c8e   blindsidenetwks/scalelite:v1.2.0.1-nginx                "/etc/nginx/start -g…"   6 minutes ago   Up 6 minutes   0.0.0.0:80->80/tcp, :::80->80/tcp, 0.0.0.0:443->443/tcp, :::443->443/tcp   scalelite-nginx
[...]

PR #701, opened 2021-10-11
Fix 404 error on recordings playback

yeganemehr:
## Description
Playback load files from /presentation/ directory, so Nginx needs to access this directory from /var/bigbluebutton/published directory.
I changed internal /static-resource location (which is useless, BTW) to /presentation to handle these requests.
as this MR merged #700 will closed.

## Testing Steps
Make a requests to `/api/getRecordings` and opened URL in `playback->format->url`.
Everything is OK now.

## Screenshots (if appropriate):
Screenshot_2021-10-11_14-46-22

### DIFF
--- a/nginx/conf.d/scalelite.common
+++ b/nginx/conf.d/scalelite.common
@@ -1,8 +1,7 @@
 # BigBlueButton recording playback
-location /static-resource {
-  alias /var/bigbluebutton/published;
+location /presentation {
+  root /var/bigbluebutton/published;
   index index.html index.htm;
-  internal;
 }
 
 location /playback/presentation/playback.html {


MrCapitalist:
> Hi Mohammad Reza! It's so basic but did you actually check that the file requested by browser is exist and accessible in nginx container? e.g if browser send a request to `https://mywebsite.com/presentation/18e94c2793d173fec3a4441bfad6eb06a3053c0a-1633939284021/metadata.xml` the file in container must be accessible in path: `/var/bigbluebutton/published/presentation/18e94c2793d173fec3a4441bfad6eb06a3053c0a-1633939284021/metadata.xml`
> 
> you can check it with this command:
> 
> ```
> docker exec -it scalelite-nginx stat /var/bigbluebutton/published/presentation/...
> ```
> 
> if the command is error you must check volumes and nfs and mount points.
> 
> if file does exist and Nginx does not serve it for you, please paste me output of this command?
> 
> ```
> docker exec -it scalelite-nginx nginx -T
> ```

I think your problem is different than mine! Because:

1. You are using v1.1.7.1 but I'm using the latest version
2. I have problem with scalelite-records-importer because my records still stay in the spool directory!

And in `nginx -t` everything is OK.

jfederico:
@yeganemehr if you do that, you are breaking the protected recording capability, which is the only modification we included in 1.2. So, not correct.

The 404 may be for other reasons, mainly a wrong configuration.

PR #720, opened 2021-10-22
GIT-718: Implement offset pagination for getRecordings (closes #718)

### LINKED ISSUE #718: Implement offset pagination for getRecordings
**Problem to solve:**

When the number of recordings reaches a fair large number, getRecordings can be difficult to handle by BBB, SL and by the integrations. There have been some workarounds on the integration side, but the need of pagination has always been there.

There is a recent initiative of implementing offset pagination in BBB, SL should match that implementation

**Purposed solution:**

Implement offset pagination for getRecordings in the form:

getRecordings?limit=20&offset=100

For a quick ref see: https://nordicapis.com/everything-you-need-to-know-about-api-pagination/

git-lama:
closes #718


kepstin:
@fcecagno @paultrudel Before we merge support for the `limit` and `offset` params into scalelite, I just want to check with you about the error codes returned for invalid values.

This current scalelite PR is adding two new messageKey values, `invalidParamLimit` and `invalidParamOffset` - but this should match what we decide to merge into BigBlueButton itself.

mabras:
Can we see this soon?
Thank you @jfederico

PR #740, opened 2021-12-07
Fix nginx startup

danielrudn:
## Description
#724 introduced an issue where nginx would fail to start with the following error:

```sh
Generating templated nginx configuration...
Using non-SSL configuration template.
Starting nginx periodic reload process...
Starting nginx...
nginx: [emerg] 91#91: unknown "scalelite_api_port" variable
nginx: [emerg] unknown "scalelite_api_port" variable
```

## Testing Steps

Try to start the container for `blindsidenetwks/scalelite:v1.3.1-beta.2-nginx`.

### DIFF
--- a/nginx/start
+++ b/nginx/start
@@ -25,11 +25,10 @@ else
 	nginx_template=/etc/nginx/conf.d/scalelite.template
 fi
 
-# apply SCALELITE_API port from environment variable or use default port 3000
-$SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}
-envsubst '$SCALELITE_API_PORT' <$nginx_template >/etc/nginx/conf.d/scalelite.conf
+SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}
 
-envsubst '$URL_HOST' <$nginx_template >/etc/nginx/conf.d/scalelite.conf
+envsubst '$URL_HOST $SCALELITE_API_PORT' < $nginx_template >/etc/nginx/conf.d/scalelite.conf
+
 unset nginx_template
 
 mkdir -p /run/nginx


nginx/start
```diff
@@ -25,11 +25,10 @@ else
...
 fi
 
-# apply SCALELITE_API port from environment variable or use default port 3000
-$SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}
-envsubst '$SCALELITE_API_PORT' <$nginx_template >/etc/nginx/conf.d/scalelite.conf
+SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}
+
+envsubst '$URL_HOST $SCALELITE_API_PORT' < $nginx_template >/etc/nginx/conf.d/scalelite.conf
 
-envsubst '$URL_HOST' <$nginx_template >/etc/nginx/conf.d/scalelite.conf
```
danielrudn, line 32: Couple issues here:

1. `$SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}` would evaluate to `3000=3000` because of shell substitution. I renamed it to `SCALELITE_API_PORT` instead.
1. `envsubst` is ran on `$nginx_template` and then immediately ran again to overwrite the previous changes. In the rendered `conf.d/scalelite.conf` the `$SCALELITE_API_PORT` variable has not be interpolated.

themad:
This still will not work, you need to export the variable to the call of `envsubst`. 
```
export SCALELITE_API_PORT=${SCALELITE_API_PORT:-3000}
envsubst '$SCALELITE_API_PORT $URL_HOST' <$nginx_template >/etc/nginx/conf.d/scalelite.conf
´´´

krim404:
this should be worked on with highest priority, as it is breaking the system

kepstin:
PR #744 obsoletes this one.

PR #772, opened 2022-04-05
Added translations to recording 404 page

farhatahmad:
## Description

NOTE: Must be followed up with instructions and changes to `docker-compose.yml`

## Testing Steps
1. Go to `http://localhost:3000/recording/1234/test`
2. Notice the page title and error messages
3. Go to `config/locales/en.yml` and change one of the strings
4. Refresh your web page and confirm the strings have changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,19 +165,19 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `PROTECTED_RECORDINGS_TOKEN_TIMEOUT`: Protected recording link token timeout in minutes. This is the amount of time that the one-time-use link returned in `getRecordings` calls will be valid for. Defaults to 60 minutes (1 hour).
 * `PROTECTED_RECORDINGS_TIMEOUT`: Protected recordings resource access cookie timeout in minutes. This is the amount of time that a user will be granted access to view a recording for after clicking on the one-time-use link. Defaults to 360 minutes (6 hours).
 * `SCALELITE_API_PORT`: Runs the SCALELITE_API in custom port number. Defaults to 3000.
-* `LOCALE`: Change the language that user facing pages displays in (currently supports `en`)
+* `DEFAULT_LOCALE`: Change the language that user facing pages displays in (currently supports `en`)
 
 ### Customizing Strings
 
-If you'd like to customize the strings on certain error pages returned by Scaltelite (`recording_not_found`), you can do so by duplicating the locale file and changing whatever lines you see fit.
+If you'd like to customize the strings on certain error pages returned by Scalelite (`recording_not_found`), you can do so by duplicating the locale file and changing whatever lines you see fit.
 
-Create the directory `/etc/default/scalelite/locales` and copy over the locale file which you want to customize the strings of. All locales can be found [here](https://github.com/blindsidenetworks/scalelite/tree/master/config/locales).
+Create the directory `/etc/default/scalelite-locales` and copy over the contents of the locales folder that can be found [here](https://github.com/blindsidenetworks/scalelite/tree/master/config/locales).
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -165,6 +165,21 @@ These variables are used by the service startup scripts in the Docker images, bu
 * `PROTECTED_RECORDINGS_TOKEN_TIMEOUT`: Protected recording link token timeout in minutes. This is the amount of time that the one-time-use link returned in `getRecordings` calls will be valid for. Defaults to 60 minutes (1 hour).
 * `PROTECTED_RECORDINGS_TIMEOUT`: Protected recordings resource access cookie timeout in minutes. This is the amount of time that a user will be granted access to view a recording for after clicking on the one-time-use link. Defaults to 360 minutes (6 hours).
 * `SCALELITE_API_PORT`: Runs the SCALELITE_API in custom port number. Defaults to 3000.
+* `LOCALE`: Change the language that user facing pages displays in (currently supports `en`)
+
+### Customizing Strings
+
+If you'd like to customize the strings on certain error pages returned by Scaltelite (`recording_not_found`), you can do so by duplicating the locale file and changing whatever lines you see fit.
```
kepstin, line 172: Typo! "Scaltelite" should be "Scalelite"

README.md
```diff
@@ -165,6 +165,21 @@ These variables are used by the service startup scripts in the Docker images, bu
...
+
+Create the directory `/etc/default/scalelite/locales` and copy over the locale file which you want to customize the strings of. All locales can be found [here](https://github.com/blindsidenetworks/scalelite/tree/master/config/locales).
```
kepstin, line 174: Perhaps we can provide a command to copy the locale file out of the docker container instead so they get the right file version? We should probably also note that it might need to be updated when new scalelite versions are release.

config/application.rb
```diff
@@ -127,5 +127,7 @@ class Application < Rails::Application
     config.x.recording_token_ttl = ENV.fetch('PROTECTED_RECORDINGS_TOKEN_TIMEOUT', '60').to_i.minutes
     # Protected recordings resource access cookie timeout in minutes. Defaults to 360 (6 hours)
     config.x.recording_cookie_ttl = ENV.fetch('PROTECTED_RECORDINGS_TIMEOUT', '360').to_i.minutes
+
+    config.i18n.default_locale = ENV.fetch('LOCALE', 'en')
```
kepstin, line 131: This variable should probably be named "DEFAULT_LOCALE" to better reflect what it does.

If multiple locales are available in the locale file, and the user's browser is set to some other locale that has strings available, then the other locale will be used instead.

config/locales/en.yml
```diff
@@ -1,33 +1,6 @@
...
-#
-# To learn more, please read the Rails Internationalization guide
-# available at https://guides.rubyonrails.org/i18n.html.
-
 en:
-  hello: "Hello world"
+  errors:
+    recording_not_found:
+      title: Recording Not Found
+      info: The recording url is invalid
```
kepstin, line 5: We can probably provide a better default string? We should probably at least mention that the recording might have been deleted, and that it might not be available because it is protected (need to use LMS links).

I recommend making the "info" string a [safe html translation](https://guides.rubyonrails.org/i18n.html#using-safe-html-translations) so that a customized locale file can use links and text formatting.

kepstin (changes requested):

README.md
farhatahmad: There's a note on the next line saying that they'll need to keep it updated. As for copying, GL recommends something like `docker run --rm bigbluebutton/greenlight:v2 cat ./sample.env > .env` in its instalation docs. Do you think something similar is good enough?

farhatahmad: We don't do that anywhere else in our docs though - so it might better to just keep it consistent

PR #780, opened 2022-05-10
GIT-781: Added env vars default and override params for create and join (closes #781)

### LINKED ISSUE #781: Extend SL to allow for overriding and appending of parameters on create and join
**Problem to solve:**

There are deployments that need to override incoming parameters from GreenLight and Moodle, such as `analytics-callback-url`.  

Instead of extending Moodle and GreenLight to include these parameters (a much longer path), we want to give them the ability to augment these parameters within ScaleLite.

**Purposed solution:**

We want to types of augmentation: override and append.  The administrator wants to configure a list of name/value pairs for each action.

For the list of *override* parameters, they will replace (or add if not provided) the corresponding parameters with the override value.

For the list of *append* parameters, they will add to the incoming parameters if they do not already exists.

There would be four lists: override and append for `create` API calls, and override and append for `join` API calls.

**Considered alternatives:**

**Additional context:**

farhatahmad:
## Description

This PR adds 4 new env variables: 
```
DEFAULT_CREATE_PARAMS
OVERRIDE_CREATE_PARAMS
DEFAULT_JOIN_PARAMS
OVERRIDE_JOIN_PARAMS
```
The `DEFAULT...` will only add the variable if it isn't already set, whereas the `OVERRIDE...` will force the param to equal whatever values are passed

## Testing Steps

1. Set any of the environment variables above using the syntax shown in the README
2. Send a create/join request
3. Confirm that the params are there as expected


jfederico (approved):

pielonet:
Please consider this issue with this PR : https://github.com/blindsidenetworks/scalelite/issues/788

PR #797, opened 2022-06-29
SL-244: Add voice bridge allocation & freeswitch dialplan generation. (closes #769)

### LINKED ISSUE #769: Add support for dial-in numbers
**Problem to solve:**
If for any reason the voice bridge fails in BBB, users can fall-back to Dial In phone number. This requires the use of a FreeSwitch server to handle the calls and a bridge between those calls and BBB.

The issue raises when clustering BBB servers behind SL. The conference numbers though need to be stored somehow in SL and linked to the `BBB Server / Meeting`

**Purposed solution:**
Extend SL to support static conference number and assign them to meetings.

**Considered alternatives:**

**Additional context:**

kepstin:
## Description
To support the ability to have POTS callers dial in to meetings handled by the scalelite load balancer, there has to be a system in place to look up the voice bridge number, and direct the call to the correct backend bigbluebutton server.

This is done by having scalelite handle allocating unique voice bridge numbers for each meeting internally, and having an index to allow looking up a meeting from its voice bridge number. Configuration options are provided to control the length of voice bridge numbers to allocate, and to choose whether or not to use voice bridge numbers provided by external integrations (as long as they're unique).

The actual dial in system itself is handled by having a freeswitch server (separate from the BBB servers) that receives external calls, and makes dialplan api calls (using mod_xml_curl) to an endpoint on scalelite. This endpoint dynamically generates dialplans that prompt for the voice bridge number and forward the call to the correct BigBlueButton server.

## Testing Steps
For the voice bridge allocation, no special test setup is required. The new configuration options are documented in the README.md, and you can create and join meetings and verify that the conference number displayed (e.g. in the welcome message) has the desired properties.

Setup for a freeswitch server for testing the dialplan and forwarding documentation is to be written.

## Notes

A lot of the "create"-related tests previously used Webmock with hard-coded URLs which no longer work when the dynamically generated voiceBridge parameter is added. These tests have all been rewritten to test for the specific parameters required on the call rather than a hardcoded URL.

Rather than work out how to do the same changes on the rspec tests, I have reimplemented the (very small number of) rspec tests in minitest and removed rspec. It's better to only have one test framework instead of 2, anyways.


app/controllers/fsapi_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      logger.info('Prompting for pin number')
+      return render 'pin_prompt'
+    end
+
+    # Look up the meeting for this voice bridge number
+    @meeting = begin
+      Meeting.find_by_voice_bridge(@pin)
+    rescue ApplicationRedisRecord::RecordNotFound
+      begin
+        # If no meeting was found, it might have been a breakout room pin, which adds one digit to the end
```
defnull, line 33: BBB supports more than 10 breakout rooms, so this logic should check both `@pin[0...-1]` and `@pin[0...-2]` or be written as a loop.

pielonet:
Hi,

could you please generate the Scalelite Docker containers for this PR so that I can test it ?

Kind regards,
Thierry

pielonet:
Hi,
when testing fsapi api I get the following error :

```
W, [2022-09-13T19:28:39.279077 #46]  WARN -- : [841da12b-77c6-4c0b-a851-5ac6608404b2] Can't verify CSRF token authenticity.
I, [2022-09-13T19:28:39.279813 #46]  INFO -- : [841da12b-77c6-4c0b-a851-5ac6608404b2] Completed 422 Unprocessable Entity in 1ms (Allocations: 254)
F, [2022-09-13T19:28:39.282420 #46] FATAL -- : [841da12b-77c6-4c0b-a851-5ac6608404b2]   
[841da12b-77c6-4c0b-a851-5ac6608404b2] ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken):
```

pielonet:
Hi,
I had it running by adding the target Location /fsapi in nginx.
Everything works great !
Thank you for this PR.
Kind regards,
Thierry

app/controllers/fsapi_controller.rb
kepstin: That's a good point. Perhaps rather than remove a fixed number of digits from the end, it would be better to try truncating the provided pin to the configured voice bridge length (if the pin is longer).

… But that wouldn't work if the "use external" option is enabled and the externally provided voice bridge numbers are shorter or longer than the configured length. Hmm.

kepstin:
> Is Generating the Docker containers for this PR as simple as :

It doesn't look like we document how to manually build the docker images right now, hmm. There's actually several docker images built using the same docker file with different "target" options. The CI system files might be the best reference for that.

Docker images for this PR branch are published to https://hub.docker.com/r/kepstin/scalelite/tags with the "dialin" prefix on the tag names.

> Hi, I had it running by adding the target Location /fsapi in nginx. 

In the nginx configuration included in the docker setup, the entire toplevel `/` is sent to scalelite (with a few exceptions), so there's no separate location block needed for `/fsapi`. I'm not sure if or where we document how to set up scalelite with an external nginx?

pielonet:
Hi @kepstin,
Thank you for providing Docker images for this PR. I well have a look !

> In the nginx configuration included in the docker setup, the entire toplevel / is sent to scalelite (with a few exceptions), so there's no separate location block needed for /fsapi. I'm not sure if or where we document how to set up scalelite with an external nginx?

I setup Scalelite using a modified version @jfederico [scalelite-run](https://github.com/jfederico/scalelite-run) installer. These scripts do not include - at the present time - a configuration for the location `/fsapi` in nginx, so I had to add it myself.

PR #807, opened 2022-08-23
Simplify recording import and increase robustness

defnull:
The current listener-based implementation is not very robust in case of temporary errors (e.g. db issues, full disc or tar files uploaded not fast enough). Those errors result in recordings sitting in the spool directory forever, even if the issue was long solved, because there is no retry-logic and existing files are not picked up again until next restart.

This patch simplifies the process by just globbing the directory every 10 seconds and re-trying failed imports every time. Listing an almost empty directory is so cheap that a 10 second interval should be fine. It's slow enough to not generate any significant IO overhead and fast enough to promptly import new recordings. Sometimes, simple is better.


git-lama (commented):
LGTM! Has it been tested?

defnull:
We have this patch applied to our fork since forever.

git-lama (approved):

lonesomewalker:
Hm... i wonder what happened to this...?

ffdixon:
Glad to see this proposal -- at the moment we're working on finishing BigBlueButton 2.6 (now in 2.6-rc.5).  We'll revisit this as part of 2.7.  Simple is good.

farhatahmad:
@defnull Thanks for the PR - have you run into any issues with this?

PR #812, opened 2022-09-09
GIT-810 & GIT-811: enforced checksum algorithms through configuration 

### LINKED ISSUE #810: Enforce the use of SHA256
**Problem to solve:**

Since [v1.3.1](https://github.com/blindsidenetworks/scalelite/releases/tag/v1.3.1), SL supports SHA256. The issue is that the implementation https://github.com/blindsidenetworks/scalelite/pull/697 allows the use of SHA1 and SHA256 together.

There are cases where the organization requires the strict use of only SHA256 (or even higher). In these cases SL should only accept the one defined.

**Purposed solution:**

Add a way to establish if SL will support multiple SHA algorithms (by default), and if it does not, only work with the one defined.

**Considered alternatives:**

**Additional context:**

jfederico:
Closes #810

### DIFF
--- a/app/controllers/concerns/api_helper.rb
+++ b/app/controllers/concerns/api_helper.rb
@@ -18,7 +18,7 @@ module ApiHelper
     raise ChecksumError unless params[:checksum].present? &&
                                (params[:checksum].length == CHECKSUM_LENGTH_SHA1 ||
                                 params[:checksum].length == CHECKSUM_LENGTH_SHA256) &&
-                                secrets.any?
+                               secrets.any?
 
     # Camel case (ex) get_meetings to getMeetings to match BBB server
     check_string = action_name.camelcase(:lower)
@@ -27,8 +27,9 @@ module ApiHelper
     )
 
     checksum_algorithms = Rails.configuration.x.loadbalancer_checksum_algorithms
-    secrets.product(checksum_algorithms).each do |secret, checksum_algorithm|
-      return if ActiveSupport::SecurityUtils.secure_compare(get_checksum(check_string + secret, checksum_algorithm), params[:checksum])
+    secrets.product(checksum_algorithms).any? do |secret, checksum_algorithm|
+      return true if ActiveSupport::SecurityUtils.secure_compare(get_checksum(check_string + secret, checksum_algorithm),
+      params[:checksum])
     end
 
     raise ChecksumError
@@ -87,7 +88,7 @@ module ApiHelper
   def post_req(uri, body)
     http = Net::HTTP.new(uri.host, uri.port)
     http.use_ssl = (uri.scheme == 'https')
-    exp = Time.now.to_i + 24 * 3600
+    exp = Time.now.to_i + (24 * 3600)
     token = encoded_token(exp: exp)
     # Setup a request and attach our JWT token
     request = Net::HTTP::Post.new(uri.request_uri,
--- a/app/controllers/concerns/api_helper.rb
+++ b/app/controllers/concerns/api_helper.rb
@@ -11,13 +11,15 @@ module ApiHelper
 
   CHECKSUM_LENGTH_SHA1 = 40
   CHECKSUM_LENGTH_SHA256 = 64
+  CHECKSUM_LENGTH_SHA512 = 128
 
   # Verify checksum
   def verify_checksum
     secrets = Rails.configuration.x.loadbalancer_secrets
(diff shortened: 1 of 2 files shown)


app/controllers/concerns/api_helper.rb
```diff
@@ -11,29 +11,38 @@ module ApiHelper
...
     check_string += request.query_string.gsub(
       /&checksum=#{params[:checksum]}|checksum=#{params[:checksum]}&|checksum=#{params[:checksum]}/, ''
     )
 
-    return if Rails.configuration.x.loadbalancer_secrets.any? do |secret|
-      checksum_sha1 = Digest::SHA1.hexdigest(check_string + secret)
-      checksum_sha256 = Digest::SHA256.hexdigest(check_string + secret)
-      ActiveSupport::SecurityUtils.secure_compare(checksum_sha1, params[:checksum]) ||
-      ActiveSupport::SecurityUtils.secure_compare(checksum_sha256, params[:checksum])
+    checksum_algorithms = Rails.configuration.x.loadbalancer_checksum_algorithms
```
kepstin, line 31: Rather than trying all supported checksum algorithms, this should try only the selected one (based on the length of the provided checksum). It still needs a check that the provided checksum matches an enabled algorithm as well, of course.

app/controllers/concerns/api_helper.rb
```diff
@@ -11,29 +11,38 @@ module ApiHelper
...
+    secrets.product(checksum_algorithms).any? do |secret, checksum_algorithm|
+      return true if ActiveSupport::SecurityUtils.secure_compare(get_checksum(check_string + secret, checksum_algorithm),
+      params[:checksum])
     end
 
     raise ChecksumError
   end
 
+  def get_checksum(check_string, checksum_algorithm)
+    return Digest::SHA512.hexdigest(check_string) if checksum_algorithm == 'SHA512'
```
kepstin, line 41: This should be reformatted as a `case` statement. Consider setting a default value for the `checksum_algorithm` option.

The method should raise an exception if an unsupported value of `checksum_algorithm` is passed.

app/controllers/concerns/api_helper.rb
```diff
@@ -82,7 +91,7 @@ def valid_token?(token)
   def post_req(uri, body)
     http = Net::HTTP.new(uri.host, uri.port)
     http.use_ssl = (uri.scheme == 'https')
-    exp = Time.now.to_i + 24 * 3600
+    exp = Time.now.to_i + (24 * 3600)
```
kepstin, line 94: Unrelated change?

config/application.rb
```diff
@@ -44,6 +44,10 @@ class Application < Rails::Application
     config.x.loadbalancer_secrets.push(ENV['LOADBALANCER_SECRET']) if ENV['LOADBALANCER_SECRET']
     config.x.loadbalancer_secrets.concat(ENV['LOADBALANCER_SECRETS'].split(':')) if ENV['LOADBALANCER_SECRETS']
 
+    # Algorithms used for calculating checksum [SHA1|SHA256|...]
+    config.x.loadbalancer_checksum_algorithm = ENV.fetch('LOADBALANCER_CHECKSUM_ALGORITHM', 'SHA1:SHA256:SHA512')
```
kepstin, line 48: Needs a documentation update in the README.md file.

kepstin (changes requested):
This work also needs some automated tests that ensure checksums from all enabled algorithms are accepted, and that checksums from disabled algorithms are rejected.

app/controllers/concerns/api_helper.rb
jfederico: Agree. Although this review should have come before. This is how the application has been dealing with multiple algorithms for more than 1 year. We may revisit this once BBB does what it is expected

app/controllers/concerns/api_helper.rb
jfederico: yeah, it would work that way too

app/controllers/concerns/api_helper.rb
jfederico: Got caught by rubocop. If it is not changed it won't pass

config/application.rb
jfederico: Correct

jfederico:
Discarded in favour of the already merged PR #828

PR #813, opened 2022-09-14
Enforce checksum algorithm

paulseto:
## Description

- refactored the selection, validation, and generating of checksums
- add support for SHA512 checksums
- add support to force the use of a specific algorithm
- updated README

## Testing Steps

1. Validated the generation of checksums using irb and php to ensure all checksums matched.
2. Used API MATE to generate urls to connect to my development environment (http://localhost:3000/...)
3. Used irb to generate urls with SHA512 checksums
4. Added ENFORCE_CHECKSUM_ALGORITHM environmental variable and retested create, join, getMeetings, getMeetingInfo actions
5. Added test cases for SHA512 checksum
6. Added test cases mock the setting of ENFORCE_CHECKSUM_ALGORITHM 
7. Generated Amazon Linux-based images and tested with Scalelite Enterprise

## Screenshots (if appropriate):

No applicable screenshots.

PR #828, opened 2022-11-02
GIT-810 & GIT-811 - force checksum algorithm through configuration + Implement support for SHA512  (closes #810) (closes #811)

### LINKED ISSUE #810: Enforce the use of SHA256

### LINKED ISSUE #811: Add support for SHA512
**Problem to solve:**
Since [v1.3.1](https://github.com/blindsidenetworks/scalelite/releases/tag/v1.3.1), SL supports SHA256. The issue is that the implementation #697 allows the use of SHA1 and SHA256 together.

But there are cases where the organization requires the strict use of only SHA384+. 

This issue is linked to #810 but blocked until it is implemented by BBB (perhaps 2.6)

**Purposed solution:**
Implement support for SHA512

borisano:
## Description

This PR finishes work started in PR #812 . 
This PR refactors the code and adds RSpec tests coverage for the APIHelper method in question, while leaving the main approach of original PR intact. 
## Testing Steps

Providing SHA512 hash will work the same way SHA1 and SHA256 currently work. 
All existing functionality should remain the same.


Gemfile.lock
```diff
@@ -87,7 +87,9 @@ GEM
     factory_bot_rails (6.2.0)
       factory_bot (~> 6.2.0)
       railties (>= 5.0.0)
-    ffi (1.15.5)
+    faker (2.22.0)
+      i18n (>= 1.8.11, < 2)
+    ffi (1.15.4)
```
jfederico, line 92: Why did we go back one version?

jfederico (changes requested):
The last pice missing is to document the new variable in the README, see 
#813 as a reference

jfederico (approved):

PR #886, opened 2023-01-20
SL-240: Implementing multitenancy

borisano:
This PR implements Multitenancy feature


pielonet:
Hi,

Thank you for this PR and the work already done to support multi-tenancy.

It looks like this PR presently only allows to create/delete tenants and secrets but not (yet) to override configuration items per tenant (like "default presentation" for example). Is that right ? When do you plan to implement this feature ?

Kind regards,
Thierry

borisano:
@pielonet 
Hi.

Custom parameters are not part of this PR, but we have this feature in a roadmap and it will be implemented soon.

jfederico (approved):

PR #911, opened 2023-04-21
Replace Minitest with RSpec: bigbluebutton_api_controller specs (1)

scouillard:
## Description

- Fix RSpec configs
- Remove `redis_helper`
- Add `bbb_api_controller` specs
- Add `Server` Factory
- Add `Meeting` Factory

## Design considerations
- Reverse the logic for checksum check: Before, we had to disable the checksum check in (almost) every test. But, considering most of the tests are not testing the checksum specifically, the checksum check is now disabled by default for all the specs and has to be enabled when testing the checksum specifically.
- In #create, a voiceBridge attribute is provided in every #create request to bypass the voiceBridge generator. This is because the stubbed request will never match the outside request as a voiceBridge generated and added to the uri by Scalelite.
- Removed the `redis_helper` file as we will be using `fakeredis` and we don't need to configure a redis database for the test environment

## Notes
-  The order of the parameters in a URL query string should not affect the functioning of the API request. For example, in #create, `override_create_params` adds the new params at the end of the uri, and `default_create_params` adds them at the beginning of the uri. Not following the order will return a different checksum value.


spec/spec_helper.rb
```diff
@@ -91,4 +98,13 @@
...
   #   Kernel.srand config.seed
+
+  config.before(:each) do
+    FactoryBot.rewind_sequences # Needed for sequences in Meeting, Server factory
+    Redis.new.flushdb
+  end
+
+  config.after(:each) do
+    Redis.new.flushdb
+  end
```
scouillard, line 109: There is a cleaner, better way to do this.
Will try to make use of `redis_helper` eventually

config/redis_store.yml
```diff
@@ -6,9 +6,10 @@ development:
   <<: *default
   namespace: <%= ENV.fetch('REDIS_NAMESPACE') { 'scalelite_development' } %>
 
+
 test:
   <<: *default
-  driver: null # Reset to allow fakeredis to load
```
scouillard, line 11: Need to discuss if we should be using fakeredis or not.

spec/factories/server.rb
```diff
@@ -0,0 +1,11 @@
...
+
+FactoryBot.define do
+  factory :server do
+    sequence(:url) { |n| "https://test-#{n}.example.com/bigbluebutton/api/" }
+    sequence(:secret) { |n| "test-#{n}-secret" }
+    load { 1 }
+    online { true }
+    enabled { true }
+  end
+end
```
scouillard, line 11: While FactoryBot is generally used with ActiveRecord objects, the current implementation of the models and redis makes it work out of the box (for now).

config/redis_store.yml
scouillard: After discussion with @jfederico , we decided it would be better not to change any testing configs to ensure the functionality of Minitest for the near future.  

Allowing the RSpec test data to be established on Redis default `db:0` requires no changes in the configs.

spec/rails_helper.rb
```diff
@@ -65,6 +69,4 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
   config.include FactoryBot::Syntax::Methods
-
-  config.include RSpec::RedisHelper, redis: true
```
scouillard, line 69: Upon reviewing, I support using `fakeredis` for our testing environment

`fakeredis` ensures that the tests are isolated from each other. As such, `redis_helper` was removed as it does not seem to be needed.

However, there might be complex Redis usage implementation in the future that makes `fakeredis` not suitable - everything seems to be working as it should on our testing environment for now.

farhatahmad (approved):

PR #917, opened 2023-04-26
Replace Minitest with RSpec: fsapi_controller specs (2)

scouillard:
## Description

Based on #911 

- Add freeswitch api controller specs
- Fix un-authenticated requests in Minitests


test/controllers/fsapi_controller_test.rb
```diff
@@ -206,7 +206,9 @@ def test_caller_id_number_hidden
 
   def test_pin_prompt_allotted_timeout_not_set
     Rails.configuration.x.stub(:fsapi_max_duration, 0) do
-      post(fsapi_url, params: { section: 'dialplan', 'Caller-Destination-Number': '5551234' })
+      FsapiController.stub_any_instance(:authenticate, true) do
+        post(fsapi_url, params: { section: 'dialplan', 'Caller-Destination-Number': '5551234' })
+      end
```
scouillard, line 211: For those two tests, the requests are not authenticated, which causes an (unauthorized) error.

Not sure if this is wanted or not, as all other test have authenticated requests.

PR #922, opened 2023-05-01
Fix Freeswitch tests - unauthenticated requests (0)

scouillard:
## Description

The last two tests in `freeswitchapi_controller_test` are failing because the requests are unauthenticated.

- Add an authenticated request stub.

Not sure if this is wanted or not, as all other test have authenticated requests.


kepstin (approved):
Fix looks correct, the tests were probably not correctly updated when the fsapi authentication support as added.

PR #928, opened 2023-05-05
Recordings size and rawSize info

blueiceprj:
## Description
Recordings size and rawSize info


blueiceprj (commented):
Recordings size and rawSize info

Issue #938, opened 2023-05-17
Multitenancy: Migration path for recordings of existing (regular) Scalelite instance to a new tenant 


Ithanil:
**Problem to solve:**

If multitenancy is enabled on an existing Scalelite deployment and configured in a way that a new tenant takes over the previous Scalelite URL and secrets 1:1, the corresponding frontends do not need to be reconfigured. Everything works as it did before the change, except for that the recordings are not available anymore.

**Purposed solution:**

Provide or document a migration path to transfer the existing recordings to the replacing tenant. Ideally, also provide a way to transfer the recordings from different Scalelites to further new tenants.

**Considered alternatives:**

Users lose all their recordings. Not ideal!

**Additional context:**

We are looking to replace multiple (at least 2) separate Scalelite deployments by one Scalelite with multi-tenancy enabled. We want to avoid losing the recordings (which are, by the way, disabled for one of the instances (for schools) and we need to be able keep it that way when using multitenancy!).

farhatahmad:
Hmm this is a good point. I'm debating whether this should be done through a rake task or manually through the console

First, get the tenant-id for the tenant that you want to move the recordings to:
```bash
docker exec -it scalelite-api /bin/bash
./bin/rake tenants
```
Then, we can update the recordings metadata:
```ruby
./bin/rails c

Recording.all.each do |rec| 
  next if rec.metadata.exists?(key: 'tenant-id')
  puts Metadatum.create(recording_id: rec.id, key: 'tenant-id', value: '4f3e4bb8-2a4e-41a6-9af8-0678c651777f')
end
```

Ithanil:
Thanks for the reply, but I won't be able to try this out until next Monday. I will report back then.

PR #942, opened 2023-05-23
SL-280: Add task to associate tenant id with all existing recordings

Mariam05:
## Description
Added a task that allows for all existing recordings to be associated with a tenant. 
Addresses issue #938

### DIFF
--- a/lib/tasks/recordings.rake
+++ b/lib/tasks/recordings.rake
@@ -42,4 +42,22 @@ namespace :recordings do
       end
     end
   end
+
+  desc 'Associate a tenant with all recordings'
+  task :import, [:tenant_id] => :environment do |_t, args|
+    tenant_id = args[:tenant_id]
+    unless tenant_id
+      puts('No tenant ID was provided')
+      exit(1)
+    end
+
+    Recording.all.each do |rec|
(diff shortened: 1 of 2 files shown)


lib/tasks/recordings.rake
```diff
@@ -42,4 +42,22 @@ namespace :recordings do
...
+  desc 'Associate any existing recordings with a tenant'
+  task :addToTenant, [:tenant_id] => :environment do |_t, args|
+    tenant_id = args[:tenant_id]
+    unless tenant_id
+      puts('No tenant ID was provided')
+      exit(1)
+    end
+
+    Recording.all.each do |rec|
+      next if rec.metadata.exists?(key: tenant_id)
```
farhatahmad, line 55: This should be `key: 'tenant-id'`

lib/tasks/recordings.rake
```diff
@@ -42,4 +42,22 @@ namespace :recordings do
...
+      begin
+        Metadatum.create!(recording_id: rec.id, key: 'tenant-id', value: tenant_id)
```
farhatahmad, line 57: Maybe we can print a success message? Either per recording or at the end just to inform the user that everything worked

farhatahmad (approved):

PR #956, opened 2023-06-01
SL-252,SL-248: Add checksum verification to servers, tenants API

Mariam05:
## Description
Changed the tenants and servers API to match the format for BBB, and added a checksum verification. 
The checksum must be computed with the LOADBALANCER_SECRET rather than the tenant's secret(s), so that tenants do not have access to all tenant and server information.

## Testing Steps
Use API Mate to create the custom calls.

### DIFF
--- a/app/controllers/api/servers_controller.rb
+++ b/app/controllers/api/servers_controller.rb
@@ -148,9 +148,9 @@ module Api
 
     def set_server
       begin
-        @server = Server.find(params[:id])
+        @server = Server.find(server_id_param[:id])
       rescue ApplicationRedisRecord::RecordNotFound
-        render json: { error: "Couldn't find server with id=#{params[:id]}" }, status: :not_found
+        render json: { error: "Couldn't find server with id=#{server_id_param[:id]}" }, status: :not_found
       end
     end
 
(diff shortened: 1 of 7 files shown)


config/routes.rb
```diff
@@ -30,14 +30,19 @@
...
+  scope 'scalelite/api', module: 'api', as: 'scalelite_api' do
+    match '/tenants', to: 'tenants#tenants', via: [:get, :post]
+    post '/tenantInfo', to: 'tenants#tenant_info', as: :tenant_info
+    post '/addTenant', to: 'tenants#add_tenant', as: :add_tenant
+    post '/updateTenant', to: 'tenants#update_tenant', as: :update_tenant
+    post '/deleteTenant', to: 'tenants#delete_tenant', as: :delete_tenant
 
-    resources :tenants, only: [:index, :show, :create, :update, :destroy]
+    match '/servers', to: 'servers#servers', via: [:get, :post]
+    post '/serverInfo', to: 'servers#server_info', as: :servers_info
```
farhatahmad, line 41: Same as above

app/controllers/api/servers_controller.rb
```diff
@@ -6,10 +6,11 @@ class ServersController < ApplicationController
 
     skip_before_action :verify_authenticity_token
 
-    before_action :set_server, only: [:show, :update, :destroy, :panic]
+    before_action :verify_lb_checksum
```
farhatahmad, line 9: I think you can drop the method and just do `before_action -> { verify_checksum(true) }`

farhatahmad (approved):

PR #965, opened 2023-06-14
Option for default presentations

paul1278:
## Description

This pull request adds the option to supply default presentations on the `create`-call. This is a good addition to the default params / override options.

When submitting a `POST`-request as `create`-call, the supplied body will be used. If there is no body supplied or it is initially a `GET`-request, the default presentations will be added.

## Testing Steps

Set the following inside your environment:
```env
DEFAULT_PRESENTATIONS: "https://test.bigbluebutton.org/default.pdf->default_1.pdf,https://test.bigbluebutton.org/default.pdf->default_2.pdf"
```
This will download the default BBB-presentation two times, with two different names to showcase the options.
`->` is the separator of URL and name because this is unlikely to appear inside an URL. Also with that, it is possible to add `base64`-presentations in the future.


pielonet:
Hi,

Thanks for this PR that we have been waiting a long time for.
Please release a beta version we can test soon.

krim404:
i would like to push this pull request, as this feature would help us a lot and we would love to see it merged in the official branch.

PR #967, opened 2023-06-15
Option for default presentations on tenants

paul1278:
## Description
This pull request adds the option to supply default presentations on the create-call on a per-tenant basis. This is a good addition to the default params / override options.

When submitting a POST-request as create-call, the supplied body will be used. If there is no body supplied or it is initially a GET-request, the default presentations for the current tenant will be added.

This PR is independent from #965. I suggest merging the other one first.
## Testing Steps
1. Add a tenant
2. Add default presentations for this tenant:
```bash
./bin/rake tenants:presentations[<id>,"https://test.bigbluebutton.org/default.pdf->default_1.pdf;https://test.bigbluebutton.org/default.pdf->default_2.pdf"]
```
3. Create a meeting and see the default presentations.

### DIFF
--- a/app/controllers/api/tenants_controller.rb
+++ b/app/controllers/api/tenants_controller.rb
@@ -84,12 +84,14 @@ module Api
     #   "id": String        # Required
     #   "tenant": {
     #     "name": String,     # include the parameter you want updated
-    #     "secrets": String
+    #     "secrets": String,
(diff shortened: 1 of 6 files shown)


pielonet:
Hi,

Thanks for this PR that we have been waiting a long time for.
Please release a beta version we can test soon.

PR #988, opened 2023-07-17
added code for custom download format (minus DB Host&PW&User)

Ithanil:
DB_HOST, DB_PASSWD, DB_USER need to be set in bin/register-download-format!

Also, we use the following nginx locations:
location ~* ^/video\/.+\/video-0.m4v {
  root /var/bigbluebutton/published;
  expires -1;
  add_header Content-Disposition "attachment; filename=video.m4v";
}

location ~* ^/video\/.+\/video-0.mp4 {
  root /var/bigbluebutton/published;
  expires -1;
  add_header Content-Disposition "attachment; filename=video.mp4";
}

### DIFF
--- a/bin/register-download-format
+++ b/bin/register-download-format
@@ -1,0 +1,63 @@
+#!/usr/bin/python3
+
+# register-download-format
+#
+# Determines test or production env from second argument or hostname,
+# opens corresponding database connection and adds Download format, if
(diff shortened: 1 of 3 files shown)


Ithanil:
Sorry, didn't want to create an upstream PR

PR #989, opened 2023-07-17
Fix bug causing getRecordings not to return all the metadata

### LINKED ISSUE #987: Multitenancy: getRecordings API call does only return tenant-id as metadata
**Describe the bug**

When Multitenancy is enabled and tenants are added, the response XML for getRecordings does only contain the tenant-id as metadata. Especially 'gl-listed' is not included, which renders all recordings "Not listed" in Greenlight.

**To Reproduce**

Enable MT with recordings. Look at recording list in GL or the response from getRecordings calls.

***Deployment:***
1. Deployment: 
2. Versions: 
3. Tools used for reproducing the issue: 
SL v1.5.1.2, custom docker-compose

**Expected behavior**

The recording metadata should be returned correctly.

**Additional context**

Uncommenting https://github.com/blindsidenetworks/scalelite/blob/2c3b29e04196a55d907b21d4737846cfa5e5f064/app/controllers/bigbluebutton_api_controller.rb#L337C17-L337C17 "fixes" the issue. I'm looking into a better way to fix this.

farhatahmad:
fixes #987 

## Description

Because of how the  query was structured, only the `tenant-id` metadata was returned, while the other metadata for the same recording was filtered out.

### DIFF
--- a/app/controllers/bigbluebutton_api_controller.rb
+++ b/app/controllers/bigbluebutton_api_controller.rb
@@ -334,7 +334,11 @@ class BigBlueButtonApiController < ApplicationController
     query = Recording.includes(playback_formats: [:thumbnails], metadata: []).references(:metadata)
 
     # Filter recordings for current tenant
-    query = query.where(metadata: { key: 'tenant-id', value: @tenant.id }) if @tenant.present?
+    if @tenant.present?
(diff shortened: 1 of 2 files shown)


Ithanil:
Can confirm it solves my issue. 👍

farhatahmad:
@Ithanil Thanks for testing it so quickly!

Ithanil:
@farhatahmad Thanks for fixing it so quickly ;-)

PR #991, opened 2023-07-19
Another take at fixing getRecordings call

farhatahmad:


farhatahmad:
@Ithanil Can you try this fix? We don't have a deployment of Scalelite that is that large, but this fix seems to work

Ithanil:
@farhatahmad Yes, it seems fine now.

Issue #1010, opened 2023-08-07
Panic does not end and clear meetings when Multitenancy is enabled


Ithanil:
**Describe the bug**

When Multitenancy is enabled, the panic rake task does not end and clear the meetings on the panicked server. According to the code, the panicServer API call should also be affected, but this is not tested. 

**To Reproduce**

Enable Multitenancy, create a meeting, panic the server. Meeting keeps running, does not get cleared, but the server will be disabled.

***Deployment:***
SL 1.5.1.5 custom docker-compose

**Expected behavior**

Panic should end all meetings on the server and clear the ended meetings, like it does when multitenancy is NOT enabled.

**Additional context**

Cause of the issue is that Meeting.all needs the tenant id as parameter to list all meetings of the tenant. Without tenant id, the list of meetings is empty.

There might be more places in the code with similar issues. 
Likely affected (aside from tests/spec):
- https://github.com/blindsidenetworks/scalelite/blob/c884b93b94f5bed296b8efd4fe5ab2e02380a01c/lib/tasks/poll.rake#L130
- https://github.com/blindsidenetworks/scalelite/blob/c884b93b94f5bed296b8efd4fe5ab2e02380a01c/lib/tasks/servers.rake#L102
- https://github.com/blindsidenetworks/scalelite/blob/c884b93b94f5bed296b8efd4fe5ab2e02380a01c/lib/server_sync.rb#L122
- https://github.com/blindsidenetworks/scalelite/blob/c884b93b94f5bed296b8efd4fe5ab2e02380a01c/lib/tasks/meetings.rake#L15

Also see https://github.com/blindsidenetworks/scalelite/issues/993

https://github.com/blindsidenetworks/scalelite/issues/889 probably unrelated.

Ithanil:
Fixed with https://github.com/blindsidenetworks/scalelite/pull/1012 / https://github.com/blindsidenetworks/scalelite/releases/tag/v1.5.1.6

Issue #993, opened 2023-07-28
Can't re-join a meeting if the previously associated server node is offline


lakano:
**Describe the bug**
Hi!
Thank you for this amazing project!
Generally we create a room few days / hours before the real live meeting, and if we go inside the meeting when we prepare it, this room seems to be definitively assigned to the selected server. 

The problem is, if at the real time where the meeting should start, this assigned server is offline, Scalelite *should* assign a new server node.

**To Reproduce**
- Create a Room
- Start the meeting, and exit/stop the meeting 
- Shutdown the node server
- Wait 1 minute ( permit to SL to mark this server node offline ) 
- Start your meeting => Can't reach the initial assigned node server

***Deployment:***
1. Deployment: docker-compose 
2. Versions: 
blindsidenetwks/scalelite   v1-focal260-alpine-poller               36f60b9b9cb6   7 days ago     1.4GB
blindsidenetwks/scalelite   v1-focal260-alpine-recording-importer   fe78ddf1b746   7 days ago     1.4GB
blindsidenetwks/scalelite   v1-focal260-alpine-api                  04e1e94dd835   7 days ago     1.4GB
blindsidenetwks/scalelite   v1-focal260-alpine-nginx                06a7d986d2ad   7 days ago     28.6MB

Ithanil:
Yes, this whole system is very broken currently. Meetings will also be assigned to disabled servers. Meetings will also remain assigned to removed servers. Thanks for creating the issue, somehow I didn't manage to make myself do it.

Ithanil:
@farhatahmad The problem is that 1) the panicking doesn't actually disassociate the meetings from the server in the Redis and b) currently - at least since MT update and with MT enabled - meetings will be placed on disabled hosts.

Try the following: Have a server online, disable it and start a meeting few seconds later. It will (maybe) be placed on the disabled host and remain on it indefinitely, leading to the following XML error:

<response>
<returncode>FAILED</returncode>
<messageKey>serverUnavailable</messageKey>
<message>The server for this meeting is disabled/offline.</message>
</response>

farhatahmad:
Yep locally atleast I can't reproduce it - will try it on one of our dev environments

lakano:
Thank you for your interest on this problem!
Yes, I've wait more than 10 minutes.
May be SL keep this original server in Redis and don't want to assign to another server after ?
If you known the Redis keys, I can investigate to see the Redis keys settings.

Ithanil:
[...]

PR #1012, opened 2023-08-07
Let Meeting.all return all meetings if tenant_id argument is not passed

Ithanil:
Let Meeting.all return all meetings, regardless of tenant association, if tenant_id argument is not passed.

## Description

Previously, with Multitenancy enabled, Meeting.all would return only meetings without tenant association if the optional argument tenant_id was not passed. This leads to numerous bugs in the code, because the tenant_id argument is not passed in many cases. An example would be https://github.com/blindsidenetworks/scalelite/issues/1010 .

I propose to let Meeting.all return all meetings, regardless of tenant association, when it is called without tenant_id. In any case where only a certain tenant is considered, the tenant_id would have to be passed anyway. This fixes the issues with meetings remaining associated after panicking/disabling servers with Multitenancy enabled.

## Testing Steps

Test all the uses of Meeting.all:

https://github.com/search?q=repo%3Ablindsidenetworks%2Fscalelite%20Meeting.all&type=code

## Fixed Issues
At least:
https://github.com/blindsidenetworks/scalelite/issues/993
https://github.com/blindsidenetworks/scalelite/issues/1010

### DIFF
--- a/app/models/meeting.rb
+++ b/app/models/meeting.rb
@@ -277,13 +277,8 @@ class Meeting < ApplicationRedisRecord
         hash = redis.hgetall(key(id))
         next if hash.blank?
 
-        if tenant_id.present?
-          # Only fetch meetings for particular Tenant
-          next if tenant_id != hash['tenant_id']
-        elsif hash['tenant_id'].present?
-          next
-        end
-        # Only fetch meetings without Tenant
+        # If tenant_id given, only fetch meetings for particular Tenant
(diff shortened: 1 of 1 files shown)


farhatahmad:
Thanks for the PR - I was looking in the wrong area

farhatahmad:
#1011 Can be closed correct?

Ithanil:
@far

> #1011 Can be closed correct?

Yes, if this PR here is merged, the other one can be closed.

Issue #1053, opened 2024-03-29
Add server groups so that different tenants can use different groups.


michaelchen001:
Hi,there
All current tenants will use all servers registered with scalelite. Is it possible to group servers so that different tenants can use all or part of the servers?

Suppose we have tenants A and B, servers S1, S2, S3, S4,
We assign the servers to Group G1(S1+S2) & G2(S1+S3+S4).
For user A, group G1 (servers S1 S2) can be used.
For user B, you can use group G2 (servers S1 S3 S4)

Ithanil:
When https://github.com/blindsidenetworks/scalelite/pull/1049 is merged, it will be possible to "abuse" the feature for the purpose tenant-specific server pools, by tagging the corresponding servers and specifying the meta_server-tag in the OVERRIDE_CREATE_PARAMS of the tenant.

Ithanil:
That said, I'm planning to make a PR with a proper implementation in the future.

PR #1066, opened 2024-05-15
Apply meta_server-tag from default/override create parameters early in create 

Ithanil:
## Description
This PR adds early handling of any meta_server-tag parameter found in DEFAULT_CREATE_PARAMS  /OVERRIDE_CREATE_PARAMS or the respective tenantSettings to the create call procedure. This is necessary to ensure the parameter has the desired effect on server selection, which occurs way before the passed create parameters are merged with configured parameters usually.
In the later occurring merging of parameters, the meta_server-tag is then excluded, to maintain the fact that the final meta_server-tag parameter is always equal to the chosen server's tag.

This PR is required to make my suggestion in https://github.com/blindsidenetworks/scalelite/issues/1053 actually work.

## Testing Steps
Tested combinations of passed vs. configured meta_server-tag via API Mate.

## Other notes
At a first glance, e.g. meta_lrs parameters would also require this treatment. I wonder if it would be better to handle all meta parameters early.


app/controllers/concerns/api_helper.rb
```diff
@@ -238,4 +241,16 @@ def add_additional_params(action, bbb_params)
...
+  # early handling of server tag in default/override create settings
+  def apply_config_server_tag(params)
+    default = Rails.configuration.x.default_create_params
+    override = Rails.configuration.x.override_create_params
+    default_tenant, override_tenant = TenantSetting.defaults_and_overrides(@tenant&.id)
+
+    params[:'meta_server-tag'] = default_tenant[:'meta_server-tag'] if params[:'meta_server-tag'].nil? && !default_tenant[:'meta_server-tag'].nil?
+    params[:'meta_server-tag'] = default[:'meta_server-tag'] if params[:'meta_server-tag'].nil? && !default[:'meta_server-tag'].nil?
+    params[:'meta_server-tag'] = override[:'meta_server-tag'] unless override[:'meta_server-tag'].nil?
+    params[:'meta_server-tag'] = override_tenant[:'meta_server-tag'] unless override_tenant[:'meta_server-tag'].nil?
```
farhatahmad, line 254: This is a bit too complex, I'm having trouble understanding / following the different cases. Can we reformat this to something that's a bit more readable with some comments?

Ithanil: If you prefer, this should be the slightly more nested, but maybe clearer, equivalent:
```
    # order of precedence: Tenant Override -> Global Override -> Param -> Tenant Default -> Global Default
    if !override_tenant[:'meta_server-tag'].nil?
      params[:'meta_server-tag'] = override_tenant[:'meta_server-tag']
    elsif !override[:'meta_server-tag'].nil?
      params[:'meta_server-tag'] = override[:'meta_server-tag']
    else
      params[:'meta_server-tag'] = default_tenant[:'meta_server-tag'] if params[:'meta_server-tag'].nil? && !default_tenant[:'meta_server-tag'].nil?
      params[:'meta_server-tag'] = default[:'meta_server-tag'] if params[:'meta_server-tag'].nil? && !default[:'meta_server-tag'].nil?
    end
 ```

 didn't try yet if Rubocop will be happy

Comment (farhatahmad) [reply app/controllers/concerns/api_helper.rb:254]: We probably simplify it quite a bit down to something closer to this
```
  params[:'meta_server-tag'] ||= 
    override_tenant[:'meta_server-tag'].presence || 
    override[:'meta_server-tag'].presence || 
    params[:'meta_server-tag'].presence || 
    default_tenant[:'meta_server-tag'].presence || 
    default[:'meta_server-tag'].presence
```

Ithanil: > We probably simplify it quite a bit down to something closer to this
> 
> ```
>   params[:'meta_server-tag'] ||= 
>     override_tenant[:'meta_server-tag'].presence || 
>     override[:'meta_server-tag'].presence || 
>     params[:'meta_server-tag'].presence || 
>     default_tenant[:'meta_server-tag'].presence || 
>     default[:'meta_server-tag'].presence
> ```

Ruby magic to the rescue, I guess. ;-) Fine!

Ithanil:
@farhatahmad Please let me test this again after your suggested change, just to be sure.

app/controllers/concerns/api_helper.rb
Ithanil: OK, the ||= assignment was not quite fine, but otherwise it works.

Ithanil:
After changing the assignment operator from ||= to = the code appears to be fine. With ||= the right hand side will not be evaluated and assigned if the left hand side is a truthy value.

PR #1129, opened 2024-12-04
Add a retry mechanism for recording ready notifier

### LINKED ISSUE #1029: Scalelite always uses LOADBALANCER_SECRET to sign recording ready notification, but frontends can be configured with one of LOADBALANCER_SECRETS instead (leading to verificatoin failure)
**Describe the bug**

If you configure one or multiple LOADBALANCER_SECRETS and use one of these as BIGBLUEBUTTON_SECRET in Greenlight (v3), GL3 fails to verify recording ready notifications, because SL will sign these using LOADBALANCER_SECRET.

Relevant code in app/services/recording_ready_notifier_service.rb:

``` 
def encoded_payload(meeting_id, record_id, tenant_name)
      secret = fetch_secrets(tenant_name: tenant_name)[0]
      payload = { meeting_id: meeting_id, record_id: record_id }
      JWT.encode(payload, secret)
end
```
In my testing  fetch_secrets(tenant_name: tenant_name)[0] returns the value of LOADBALANCER_SECRET.

Related issue in GL: https://github.com/bigbluebutton/greenlight/issues/5392

***Deployment:***
SL + GLv3

**Expected behavior**

As long as one is allowed to use one of the LOADBALANCER_SECRETS as opposed to the LOADBALANCER_SECRET for a frontend, SL needs to use the correct secret for the recording ready notification or not encode this at all.

farhatahmad:
fixes #1029

## Description

Add a "retry" mechanism to loop through all of the loadbalancer secrets when sending the recording ready notification service

Greenlight PR: https://github.com/bigbluebutton/greenlight/pull/5972

### DIFF
--- a/app/services/recording_ready_notifier_service.rb
+++ b/app/services/recording_ready_notifier_service.rb
@@ -21,8 +21,7 @@ class RecordingReadyNotifierService
       notify(callback_url, meeting_id, recording.record_id, tenant_name) if callback_url
     end
 
-    def encoded_payload(meeting_id, record_id, tenant_name)
-      secret = fetch_secrets(tenant_name: tenant_name)[0]
+    def encoded_payload(meeting_id, record_id, secret)
       payload = { meeting_id: meeting_id, record_id: record_id }
       JWT.encode(payload, secret)
     end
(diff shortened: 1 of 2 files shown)


Ithanil (approved):
LGTM

I mean, this kind of brute forcing is not the prettiest solution, but doing it "properly" would be considerably more effort.
I have also successfully tested the changes in actual deployment with a Scalelite and a (correspondingly patched) Greenlight, where one of LOADBALANCER_SECRETS is used for the GL.

PR #1133, opened 2024-12-12
Speed up docker build

defnull:
Use official ruby base images instead of compiling ruby from source


defnull:
No idea why there are two identical Dockerfiles in this repo

Ithanil:
I have no idea either. Greenlight repo has the same structure, but there the files are different. 🤷🏼‍♂️

PR #1151, opened 2025-04-17
Fix Anayltic Callback Issues

farhatahmad:
## Description

Initial attempt to fix Analytic Callback Issues when multiple `LOADBALANCER_SECRETS` are defined


antobinary:
cc @soheb-somani

Issue #1047, opened 2024-02-06
`scalelite_post_publish.rb` ignores `-f` parameter and transfers formats multiple times


defnull:
**Describe the bug**
Recording post-publish scripts are called once for each format generated by BBB, but the `scalelite_post_publish.rb` script ignores the `-f` parameter and transfers all available formats for a given recording every time it is called. If BBB is configured to produce more than one formats (e.g. presentation, video and podcast), the script will be called multiple times and transfer the early formats multiple times, resulting in multiple imports of the same format by the Scalelite importer. The importer just replaces existing imports, so this issue only produces unnecessary work but may go unnoticed otherwise.

**To Reproduce**
Just look at https://github.com/blindsidenetworks/scalelite/blob/7f0cbde4d934c983a2e5c3175a2a7b376bd1b540/bigbluebutton/scalelite_post_publish.rb#L32

**Expected behavior**
The `scalelite_post_publish.rb` script should only pack and transfer the format specified by the `-f` parameter.

defnull:
Oh, according to #888 this is more serious than I thought and causes data loss if `delete_recording: true` is enabled. The script is not prepared to handle multiple recording formats.

PR #1247, opened 2026-04-30
Add support for waiting until all recording formats are published

michelleDeko:
## Description
This PR adds a new option called `wait_for_all_formats` to the BBB post_publish script.

If this option is enabled, the post_publish script will wait until all formats, that are active in `/usr/local/bigbluebutton/core/scripts/publish/*.rb`, are published.

I added this since we sometimes had a case, where the video, notes or our own format was just missing. This should also reduce network traffic since the transfer happens only once now.

## Testing Steps

- Deploy the new post_publish script
- Enable option `wait_for_all_formats` in scalelite.yml
- Create a recording
- Inside the bbb-rap-resque-worker logs you will now see: `Skipping transfer - waiting for all formats to finish (triggered by format: video)`


michelleDeko:
The new commit now also fixes https://github.com/blindsidenetworks/scalelite/issues/1047

farhatahmad:
I spoke to Calvin about this - unfortunately this doesn't entirely work. The notes format does not create a `.done` file if the shared notes wasn't used, which would result in this hanging for a while. 

We spoke about potentially adding another hook when all recordings are published, but for now, I think it's best for Scalelite to continue working as is until we can find a suitable workaround

michelleDeko:
Hi @farhatahmad!

I don't think this is true, just confirmed it myself, created a 3 second long meeting with no notes, only drew something on the whiteboard and:
```
root@bbb-xxxxxx:/var/bigbluebutton/recording/status/processed# ls -la | grep 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789
-rw-r--r-- 1 bigbluebutton bigbluebutton    64 Jul  1 11:29 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789-notes.done
-rw-r--r-- 1 bigbluebutton bigbluebutton    64 Jul  1 11:29 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789-presentation.done
-rw-r--r-- 1 bigbluebutton bigbluebutton    64 Jul  1 11:29 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789-summary.done
-rw-r--r-- 1 bigbluebutton bigbluebutton    64 Jul  1 11:29 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789-transcript.done
-rw-r--r-- 1 bigbluebutton bigbluebutton    64 Jul  1 11:29 296251c968cf4c9db33bbd614ee67a1e4fa906a2-1782898145789-video.done
```

We also already use the `wait_for_all_formats` option and the commit which sends recording formats in a seperate tar file in production and the notes always create a .done file, even when notes weren't used.

PR #1259, opened 2026-06-18
Implement graceful shutdown for recording importer on SIGTERM/SIGINT

michelleDeko:
## Description
Currently the scalelite-recording-importer gets instantly shutdown, even when there are still running imports.
With this PR, a running import can finish before the scalelite-recording-importer gets shutdown.
Depending on how you run scalelite, a timeout can be added with `--stop-timeout 120` or:
```
services:
  scalelite-recording-importer:
    stop_grace_period: 2m
```

## Testing Steps
- Create Meeting with recording, end it and wait until scalelite imports it.
- Turn off the container
- Recording should still get imported


farhatahmad:
Thanks for PR Michelle - can you also make the change to the systemd files that would be need to make this work?

michelleDeko:
Hi Ahmad!

The change to the systemd service file has been made

